PR #10, opened 2014-04-07
Add Data URI (RFC 2397) Wagon provider

jycr:
Add Data URI (RFC 2397) Wagon provider


michael-o:
Is there any real world example for this?

jycr:
I use this to easely encapsulate Version number rules (to exclude alpha, beta and RC version)
(cf. http://www.mojohaus.org/versions-maven-plugin/version-rules.html)

michael-o:
I'd be interested to see an IT with your usecase.

jycr:
I've added some documentations around how to use this feature.

Can you give me an example of IT (in a representative project/module) so I can see how to do?

michael-o:
I guess you are referring to `index.apt.vm`. It looks really funny and also light a maintenance nightmare. Does this really make sense? Wouldn't it be easier to put it under `src` somewhere?

> Can you give me an example of IT (in a representative project/module) so I can see how to do?

I'll provide one when we have clarified.

jycr:
To be transparent with you, I've created this extension because I have not found an effective way to share (in my company and all the projects it manages), a rule to exclude non-stable versions (beta, alpha, RC, etc) of dependencies.

I have to share this rule with our suppliers.
The only thing I can give to them is a "master POM" which they have to inherit.

Unfortunately I do not have the possibility to provide, via a public URL, this rule (to exclude non-stable versions).

I did not find any other way to configure the behavior of versions-maven-plugin to achieve this result.

I totally agree with you on the fact that the opacity of the data URI is difficult to maintain, but now I have not found a better solution.

If you think that this extension is not a good thing, do not waste time, you can simply refuse it.

michael-o:
@jycr I absolutely do not intend to deny your efforts but currently I do not see a broad public use for this. The solution seems quiet limited to your situation. I am glad that you where able to solve your problem.

There is no way for your company to provide a TLS-protected public URL for that data?

PR #16, opened 2014-11-26
Prevent fingerprints loss in known_hosts

marob:
Currently, when adding new fingerprints the whole known_hosts file is rewritten from the loaded fingerprints.
But only fingerprints with a compatible algorithm (from jsch point of view) are kept: for example, all "ecdsa-sha2-nistp256" are lost!

Hence this pull request that appends the new fingerprint in the known_hosts file (without modifying the rest of the file) and keeps unknown fingerprints with unknown algorithms.

### DIFF
--- a/wagon-providers/wagon-ssh-common/src/main/java/org/apache/maven/wagon/providers/ssh/knownhost/AbstractKnownHostsProvider.java
+++ b/wagon-providers/wagon-ssh-common/src/main/java/org/apache/maven/wagon/providers/ssh/knownhost/AbstractKnownHostsProvider.java
@@ -66,5 +66,8 @@ public abstract class AbstractKnownHostsProvider
     {
     }
     
-    
+    public void addKnownHost( KnownHostEntry knownHost )
+        throws IOException
+    {
+    }
 }
--- a/wagon-providers/wagon-ssh-common/src/main/java/org/apache/maven/wagon/providers/ssh/knownhost/FileKnownHostsProvider.java
+++ b/wagon-providers/wagon-ssh-common/src/main/java/org/apache/maven/wagon/providers/ssh/knownhost/FileKnownHostsProvider.java
@@ -81,6 +81,16 @@ public class FileKnownHostsProvider
         }
     }
     
+    public void addKnownHost( KnownHostEntry knownHostEntry )
+        throws IOException
+    {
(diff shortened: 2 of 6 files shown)


didlich:
see also the duplicate: http://jira.codehaus.org/browse/WAGON-436

I vote for this PR
+1

marob:
@olamy @hboutemy Please, consider this PR

didlich:
@brettporter @carlossg @olamy @hboutemy I noticed a new release was published recently, is there a chance the described problem here is already solved in version 2.9?

didlich:
problem remains with version 2.9, it is still killing my known_hosts file on the build server
I am deeply saddened :-(

marob:
It's been months... can we expect any merge one day?

didlich:
+1

dantran:
I will take a look. Please bare with me on my ignorance, and provide me a test case, promise to push 2.10.1 out for you. Even a pom using wagon-wagon-plugin also fine

marob:
As you can see, my modifications are in the AbstractJschWagon.openConnectionInternal method. This method launches a real connection and cannot be easily tested.
For example, it contains the following line (that instantiates a real JSch instance that is not mockable):
```java
JSch sch = new JSch();
```

Some important refactoring would be required to make it testable.

Nevertheless, I've built a POC at https://github.com/marob/maven-wagon-426 to demonstrate the bug and my fix.
Please, take the time to test my POC.

Regards

dantran:
Just want confirm, only the addional line that inserted prior to your test get removed, or the entire file is empty out

marob:
All lines with non recognized algorithms are deleted. So the newly added line and all other ecdsa fingerprints (if any).
Rsa and dsa fingerprints are kept, with our without my fix.

dantran:
I had to revert the change, its NPE wagon-ssh at wagon-maven-plugin/src/it/ssh-it test.

dantran:
hkr.getHostKey( host, null ) can return null where to orignal code does handle

dantran:
it is fixed and test with wagon-maven-plugin. Thanks

dantran:
please test wagon-2.11-SNAPSHOT

marob:
OK. I've tested 2.11-SNAPSHOT and can confirm the FIX is OK.
Thanks!

PR #18, opened 2014-12-23
Implemented mkdirs method to use the webdav MKCOL HTTP verb to create directories

nigelsim:
This implements the mkdirs method so that the HTTP Wagon can function as a replacement for the Jackrabbit WebDav Wagon, which is an advantage as the HttpClient 4.x supports SNI and we also have better control over SSL handling.


michael-o:
Is this a complete replacement for Jackrabbit for the Wagon provider?

michael-o:
One could also consider merging both and differentiate only by the URL, if this is possible.

michael-o:
Any response?

michael-o:
The `mkdirs()` works well, just tested.

PR #19, opened 2015-04-17
[WAGON-432] make sure User-Agent header is sent exactly once

jsievers:
- remove unnecessary explicit default User-Agent (all 3 HTTP wagon
  providers [http, lightweight, webdav] send a User-Agent by default
  already)
- when adding headers, use setHeader() instead of addHeader().
  this will avoid sending the same header key multiple times
- adapt header listener in tests so it will detect header keys
  sent multiple times
- add tests to make sure default User-Agent is always present and
  User-Agent is present only once even if set multiple times
  using different methods

Signed-off-by: Jan Sievers <jan.sievers@sap.com>


jsievers:
thanks!

PR #23, opened 2016-02-17
[WAGON-451] Accept NO CONTENT in resourceExists

hyPiRion:
Some maven repositories may respond with 204 NO CONTENT instead of 200 OK when sending in HTTP HEAD requests. This throws an exception in wagon-http and wagon-http-lightweight.

This patch makes 204 NO CONTENT responses from HTTP HEAD requests in wagon-http and wagon-http-lightweight return true in `resourceExists`, instead of throwing an exception.

The patch also adds makes wagon-http-lightweight accept 304 NOT MODIFIED, to make it consistent with the wagon-http implementation.

### DIFF
--- a/wagon-providers/wagon-http-lightweight/src/main/java/org/apache/maven/wagon/providers/http/LightweightHttpWagon.java
+++ b/wagon-providers/wagon-http-lightweight/src/main/java/org/apache/maven/wagon/providers/http/LightweightHttpWagon.java
@@ -392,7 +392,9 @@ public class LightweightHttpWagon
 
             switch ( statusCode )
             {
-                case HttpURLConnection.HTTP_OK:
+                case HttpURLConnection.HTTP_OK: // 200
+                case HttpURLConnection.HTTP_NO_CONTENT: // 204
+                case HttpURLConnection.HTTP_NOT_MODIFIED: // 304
                     return true;
 
                 case HttpURLConnection.HTTP_FORBIDDEN:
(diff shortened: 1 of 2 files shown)


michael-o:
> Some maven repositories may respond with 204 NO CONTENT instead of 200 OK when sending in HTTP HEAD requests.

If so, they are violating the RFCs. File an issue against those repos.

hyPiRion:
> > Some maven repositories may respond with 204 NO CONTENT instead of 200 OK when sending in HTTP HEAD requests.
> 
> If so, they are violating the RFCs. File an issue against those repos.

Yeah, that was what I originally thought too, but I noticed Apache Ivy [fixed the exact same issue](https://github.com/apache/ant-ivy/commit/899f9fd0a9690581f33a715bb5f000397f1bafef.patch) some years ago. So I went to look at the actual specification, and the RFC section on [HEAD](http://tools.ietf.org/html/rfc7231#section-4.3.2) seems to allow it, _technically_ speaking.

I agree that it does not make sense though, so feel free to close this PR if you folks feel they are violating the RFC, and I'll bring it back to the repo maintainers.

michael-o:
If I were at Ivy, I would have reject the PR at all. It is silly to workaround such isues if you can fix the root cause. What repo managers are we talking about? I am inclined to reject this PR and the JIRA issue for this.

michael-o:
If I were at Ivy, I would have reject the PR at all. It is silly to workaround such isues if you can fix the root cause. What repo managers are we talking about? I am inclined to reject this PR and the JIRA issue for this.

hyPiRion:
Alright, sorry for the disturbance.

michael-o:
It wasn't one because we added documentation benefit to the issue.

PR #25, opened 2016-07-29
add extension for external tool

soiff:
Add extension for external download tool such as axel.


soiff:
fix defect in new source

PR #27, opened 2016-07-29
Support external downloader for downloading files

soiff:
add extension for external downloader by specifying following properties:

```
external.loader=axel
external.filter=.*\\.jar
external.state=.st
```

which means one can download all jar resource through `axel`.

### DIFF
--- a/wagon-provider-api/src/main/java/org/apache/maven/wagon/StreamWagon.java
+++ b/wagon-provider-api/src/main/java/org/apache/maven/wagon/StreamWagon.java
@@ -89,7 +89,7 @@ public abstract class StreamWagon
 
         if ( new StreamWagonEx( this ).handle( timestamp, resource, destination ) )
         {
-            IOUtil.close(is);
+            IOUtil.close( is );
 
             return true;
         }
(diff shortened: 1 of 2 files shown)


soiff:
add default filter

michael-o:
Why is `wagon-http` and `wagon-http-lightweight` not enough?

soiff:
It's not enough for unstable network connections and shows not enough log about downloadings, suppose we are behind the GFW. By the way, support external downloaders make wagon itself more flexible. Of course, this patch do not force users to adopt it, which is optional.

michael-o:
If it is not enough for unstable networks, provide PRs to make it bullet-proof. Logs can always we enabled with `-X`  and by tweaking the logging configuration. Still not enough?

soiff:
Yes, you are right. There are no PRs to make it bullet-proof. I've encountered this problem several days ago and I tried this new patch, which solved my problem and it works fine. In my memory, I've blocked for many times when I try to get new releases for artifacts from central repository and it's just waiting without any information print out, which makes me nervous and incomfortable. Of course, nowadays, we use local repository for maven instead of central repository, which reduce this problem too. From my perspective, I just want to make wagon more flexible for users, nothing else.

michael-o:
I do agree, but still think that we can tweak Wagon HTTP to fail faster if such a case happens. Your proxy is likely to use Apache HttpClient, just like Wagon does. You could still suffer from the same problems. I really would like to see this flexibility in a well-established provider which is the default, rather than people having to declare and configure a new one.

soiff:
Actually, `axel` is based on `c` programing language and do really failfast I think. Of course this is not a problem that always happen, which means it would be narrowly needed. After all, improving the default provider should be the more acceptible way. So I'm going to close this comment and post any sugestion if I encountered such problems again. Thanks for your attention!

PR #28, opened 2016-10-27
Add support for detecting and using of KiTTY's klink.exe and kscp.exe

sebthom:
KiTTY is a PuTTY fork available at http://www.9bis.net/kitty/

When using KiTTY's klink.exe/kscp.exe instead of PuTTY's plink.exe/pscp.exe, the ScpExternalWagon implementation passes the command line arguments "-o BatchMode yes" to these executables which fail with "unknown option -o".

This patch changes the ScpExternalWagon class to treat KiTTY executables like PuTTY executables.


michael-o:
Please create a JIRA issue for that an I will merge the PR.

sebthom:
At last: https://issues.apache.org/jira/browse/WAGON-525

PR #31, opened 2016-12-09
[WAGON-469] wagon-http does not enable Expect-Continue by default

mcculls:
There appear to be two issues:
* Expect-Continue is not enabled for PUT when no configuration is provided
* The default Expect-Continue configuration parameter uses the wrong key


michael-o:
Looks very good. Please squash and I will merge it.

mcculls:
OK, squashed

PR #32, opened 2017-01-09
[WAGON-485] ScpWagon file size Integer to Long

oassuncao:
Change filesize to Long as `resource.setContentLength`

### DIFF
--- a/wagon-providers/wagon-ssh/src/main/java/org/apache/maven/wagon/providers/ssh/jsch/ScpWagon.java
+++ b/wagon-providers/wagon-ssh/src/main/java/org/apache/maven/wagon/providers/ssh/jsch/ScpWagon.java
@@ -297,7 +297,7 @@ public class ScpWagon
                 throw new IOException( "Invalid transfer header: " + line );
             }
 
-            int filesize = Integer.valueOf( line.substring( 5, index ) ).intValue();
+            long filesize = Long.parseLong( line.substring( 5, index ) );
             fireTransferDebug( "Remote file size: " + filesize );
 
             resource.setContentLength( filesize );


michael-o:
Is the value in bytes? If though, this truly can fail with intergers. Please open a JIRA issue for that.

oassuncao:
If you check the `resource.setContentLength` this method receive a long value as a parameter. In the `ScpWagon` convert the integer to long (java cast). 

I only parse the value to Long, because if you have large file you will receive a error. My file has 3.8 GB in bytes is 3865470566. It is more than the capability of Integer Java

wagon-providers/wagon-ssh/src/main/java/org/apache/maven/wagon/providers/ssh/jsch/ScpWagon.java
```diff
@@ -297,7 +297,7 @@ public void fillInputData( InputData inputData )
                 throw new IOException( "Invalid transfer header: " + line );
             }
 
-            int filesize = Integer.valueOf( line.substring( 5, index ) ).intValue();
+            long filesize = Long.valueOf( line.substring( 5, index ) );
```
Tunaki: Small note, `Long.parseLong` is better here as it wouldn't involve an unboxing conversion.

michael-o: Good point. Waiting for this to be integrated and I will ran all tests against the patch.

oassuncao:
Changed the code to use `Long.parseLong`

michael-o:
Please squash them.

oassuncao:
I squash the commits.

Thanks

stephenc:
LGTM

jira-importer:
Resolve #541

PR #33, opened 2017-03-22
WAGON-363 wagon-ftp omit duplicate CWD

lbreuss:
I incorporated the suggested patch and built it inhouse at our company, as version 2.12-WAGON-363. It resolves the problem. (The logic in the code could be restructured a bit to be more readable, but it does the job)


jira-importer:
Resolve #425

PR #35, opened 2017-10-13
new feature: support stream read interface

ileler:


elharo:
This would need to be discussed on the dev mailing list and in JIRA first.

PR #36, opened 2017-11-07
[WAGON-452] RelaxedTrustStrategy handle multiple certificates

DonnyDepp:
Ignoring ssl validity of dates is working, but in my case the server
returns 3 (expired) certificates.
With that RelaxedTrustStrategy always returned false.

When RelaxedTrustStrategy returns false, then trustManager.checkServerTrusted(...) runs into the CertificateExpiredException, so"maven.wagon.http.ssl.ignore.validity.dates" becomes obsolete.

The fix is to check all existing certificates in the given way.


jan-z:
I can confirm this patch fixes the problem. Please merge.

jira-importer:
Resolve #509

PR #40, opened 2018-02-18
[WAGON-497] fix ScmWagon.put() strips existing parent directories

basinilya:


olamy:
can please squash your commits as well.

basinilya:
done

PR #44, opened 2018-02-21
[WAGON-500] re-enable ScmCvsExeWagonTest

basinilya:


olamy:
thanks but do we really want spend (waste?) time for cvs :)

basinilya:
At least apply the patches to the java files.

basinilya:
There's a quick and a proper way. The quick way is to return from test methods when no cvs binary is found, just like maven-scm does. I can push the change today.

 The proper way is to migrate to PlexusJUnit4TestCase and use the Assume.assumeTrue() which marks the test as skipped in runtime.

michael-o:
I take either. You decide what you want to provide.

basinilya:
Pushed. Please check.

PR #49, opened 2018-08-17
[MNG-2802] Concurrent-safe access to local Maven repository

erikhakansson:
Made file wagon optionally lock files using an adapted version of LockableFileWriter from commons-io.

Integration test is made by forking two separate java processes that attempts to lock the same file. Only the first will succeed.

At first I tried using Java's FileLock, but that proved impossible on Windows Subsystem for Linux as that doesn't reliably support file locks, and after some googling I realized the native FileLocks are generally unreliable from Java. Instead I opted to use lockfiles using the commons-io approach.

Use system property maven.wagon.file.fileLock=true to enable file lock. Default is off, default timeout is 300 seconds, but configurable with maven.wagon.file.fileLock.timeout

### DIFF
--- a/wagon-providers/wagon-file-test/src/main/java/org/apache/maven/wagon/providers/file/LazyLockableFileOutputStreamTestDriver.java
+++ b/wagon-providers/wagon-file-test/src/main/java/org/apache/maven/wagon/providers/file/LazyLockableFileOutputStreamTestDriver.java
@@ -1,0 +1,75 @@
+package org.apache.maven.wagon.providers.file;
+
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+import java.io.File;
+import java.io.IOException;
+import java.io.OutputStream;
+import java.util.concurrent.TimeUnit;
+
+/**
+ * Test driver for LazyLockableFileOutputStreamIntegrationTest
+ */
+public class LazyLockableFileOutputStreamTestDriver
+{
+    public static void main( String[] args ) throws Exception
+    {
+        File file = new File( args[0] );
+        long timeout = Long.parseLong( args[1] );
+        long startTime = System.currentTimeMillis();
+        OutputStream outputStream = new LazyLockableFileOutputStream( file, timeout, TimeUnit.SECONDS );
+        try
+        {
+            outputStream.write( 1 );
+        }
+        catch ( IOException e )
+        {
+            if ( e.getMessage().equals( "Can't write file, lock " + file.getAbsolutePath() + ".lck exists" ) )
+            {
+                System.out.println( "ready" );
+                // CHECKSTYLE_OFF: MagicNumber
+                System.exit( 126 );
+                // CHECKSTYLE_ON: MagicNumber
+            }
+            else if ( e.getMessage().equals( "Failed to create lockfile " + file.getAbsolutePath()
+                    + ".lck after waiting " + timeout + " seconds. File already exists." ) )
+            {
+                long diff = System.currentTimeMillis() - startTime;
+                if ( diff < TimeUnit.SECONDS.toMillis( timeout ) )
+                {
+                    throw new Exception( "We were supposed to wait for " + timeout
+                            + " seconds, but Exception came early at " + diff + " milliseconds." );
+                }
+                System.out.println( "ready" );
+                // CHECKSTYLE_OFF: MagicNumber
+                System.exit( 127 );
+                // CHECKSTYLE_ON: MagicNumber
+            }
+            else
+            {
+                throw e;
+            }
+        }
+        System.out.println( "ready" );
+        //noinspection ResultOfMethodCallIgnored
+        System.in.read(); //wait for input to allow test to control when to exit.
+        outputStream.close();
+    }
+}
--- a/wagon-providers/wagon-file/src/main/java/org/apache/maven/wagon/providers/file/WaitingLockableFileWriter.java
+++ b/wagon-providers/wagon-file/src/main/java/org/apache/maven/wagon/providers/file/WaitingLockableFileWriter.java
@@ -1,0 +1,301 @@
+package org.apache.maven.wagon.providers.file;
+
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+import org.apache.commons.io.Charsets;
+import org.apache.commons.io.FileUtils;
+
+import java.io.File;
+import java.io.FileOutputStream;
+import java.io.IOException;
+import java.io.OutputStreamWriter;
+import java.io.Writer;
+import java.nio.charset.Charset;
+import java.util.concurrent.TimeUnit;
+
+/**
+ * CHECKSTYLE_OFF: LineLength
+ * Adapted from commons-io LockableFileWriter
+ * https://github.com/apache/commons-io/blob/58b0f795b31482daa6bb5473a8b2c398e029f5fb/src/main/java/org/apache/commons/io/output/LockableFileWriter.java
+ * licenced under apache-2.0
+ *
+ * Aside from some clean-up of not used parts, the main difference is that we can optionally wait for a file lock.
+ *
+ * @author <a href="mailto:erikhakan@gmail.com">Erik Håkansson</a>
+ * CHECKSTYLE_ON: LineLength
+ */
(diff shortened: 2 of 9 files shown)


wagon-providers/wagon-file-test/src/main/java/org/apache/maven/wagon/providers/file/LazyLockableFileOutputStreamTestDriver.java
```diff
@@ -0,0 +1,75 @@
...
+            {
+                long diff = System.currentTimeMillis() - startTime;
+                if ( diff < TimeUnit.SECONDS.toMillis( timeout ) )
+                {
+                    throw new Exception( "We were supposed to wait for " + timeout
+                            + " seconds, but Exception came early at " + diff + " milliseconds." );
+                }
+                System.out.println( "ready" );
+                // CHECKSTYLE_OFF: MagicNumber
+                System.exit( 127 );
```
olamy: please avoid System.exit this can fail to correctly report failures with surefire

wagon-providers/wagon-file-test/src/main/java/org/apache/maven/wagon/providers/file/LazyLockableFileOutputStreamTestDriver.java
```diff
@@ -0,0 +1,75 @@
...
+        {
+            outputStream.write( 1 );
+        }
+        catch ( IOException e )
+        {
+            if ( e.getMessage().equals( "Can't write file, lock " + file.getAbsolutePath() + ".lck exists" ) )
+            {
+                System.out.println( "ready" );
+                // CHECKSTYLE_OFF: MagicNumber
+                System.exit( 126 );
```
olamy: please avoid System.exit this can fail to correctly report failures with surefire

wagon-providers/wagon-file-test/src/main/java/org/apache/maven/wagon/providers/file/LazyLockableFileOutputStreamTestDriver.java
erikhakansson: Ah, I was not aware of that! But since the LazyLockableFileOutputStreamTestDriver isn't directly managed by surefire, but instead launched in a separate JVM, shouldn't it be quite safe anyway?

It's a very convenient way of reporting results out to the actual test without having to read the output stream. Which would be the alternative solution.

erikhakansson: ping @olamy 
what do you think?

erikhakansson: @olamy sorry for nagging, but have you had time to think about this?

erikhakansson:
Can someone have a new look at this after my comments regarding System.exit?

To reiterate, the System.exit isn't run in anything manaded by Surefire and thus shouldn't affect/be affected by surefire behaviour. LazyLockableFileOutputStreamTestDriver is never run by Surefire

The test (LazyLockableFileOutputStreamIntegrationTest) launches two separate JVMs to be able to properly isolate file locking behaviour and not having different processes in the same JVM affect each other. I.e, simulating real world behaviour with different instances of Maven accessing the same file.

michael-o:
Looking into...

wagon-providers/wagon-file/src/main/java/org/apache/maven/wagon/providers/file/WaitingLockableFileWriter.java
```diff
@@ -0,0 +1,301 @@
...
+                while ( lockFile.exists() )
+                {
+                    if ( System.currentTimeMillis() - startTime >= timeoutInMilliseconds )
+                    {
+                        throw new IOException( "Failed to create lockfile " + lockFile + " after waiting "
+                                + timeoutUnit.toSeconds( timeout ) + " seconds. File already exists." );
+                    }
+                    try
+                    {
+                        Thread.sleep( 512 );
```
michael-o: OK, I now see that you wait in chunks and give up in at most `timeout`.

wagon-providers/wagon-file/src/main/java/org/apache/maven/wagon/providers/file/WaitingLockableFileWriter.java
```diff
@@ -0,0 +1,301 @@
...
+ * Adapted from commons-io LockableFileWriter
+ * https://github.com/apache/commons-io/blob/58b0f795b31482daa6bb5473a8b2c398e029f5fb/src/main/java/org/apache/commons/io/output/LockableFileWriter.java
+ * licenced under apache-2.0
+ *
+ * Aside from some clean-up of not used parts, the main difference is that we can optionally wait for a file lock.
+ *
+ * @author <a href="mailto:erikhakan@gmail.com">Erik Håkansson</a>
+ * CHECKSTYLE_ON: LineLength
+ */
+public class WaitingLockableFileWriter extends Writer
```
michael-o: The `LockableFileWriter` from Commons IO is a good idea, but not really suited for this IMHO, because Wagon transports opaque binary files and not text files what I writer is for. Am I mistaken?

erikhakansson:
Hello!

It's been a long time since I had the ideas of this PR fresh in my mind, but do you think it would be worth implementing the same basic ideas in `BasicRepositoryConnector` instead? I have had zero look at that before, though, and it might be a while before I have time to check up on it.

You may very well be right about the LockableFileWriter now that I think about it, but the same principle should work fine with a BAOS and FileOutputStream or something.

I understand your dislike for the System.exit() calls, but they are a convenient way to report status back to owning process, but of course some sort of stdout could be utilized instead. They shouldn't cause any harm though since whatever happens in the spawned process should be invisible to the containing JVM.

michael-o:
> It's been a long time since I had the ideas of this PR fresh in my mind, but do you think it would be worth implementing the same basic ideas in BasicRepositoryConnector instead? I have had zero look at that before, though, and it might be a while before I have time to check up on it.

I think it is. You should take a look at `BasicRepositoryConnector` and tell what you think and how you could change that because putting anymore effort into it. If we can agree to some reasonable implementation you can take the take and update your PR.

> You may very well be right about the LockableFileWriter now that I think about it, but the same principle should work fine with a BAOS and FileOutputStream or something.

Correct, a `LockableFileOutputStream` might work as well. BAOS not because we could have gigabyte large files.

> I understand your dislike for the System.exit() calls, but they are a convenient way to report status back to owning process, but of course some sort of stdout could be utilized instead. They shouldn't cause any harm though since whatever happens in the spawned process should be invisible to the containing JVM.

I think this is our least problem, I noticed that they are scoped in the forked process so they cannot drown the Surefire VM.

erikhakansson:
@michael-o, after looking at `maven-resolver` a bit, it seems that there's already file-locking in place during download (in `PartialFile`). It locks on the tempfile though, but since the tempfile is a fixed known name it seems unlikely that different Maven instances should write to different tempfiles.

I.e, maven instance 2 should see the file lock from instance 1.

However, there's a tiny window where instance 2 might start checking just before the target file is created but doesn't reach the actual file locking part until process 1 has already finished.

However, what confuses me is that at worst case this should just trigger a re-download. The final move from tempfile to target file (in `GetTaskRunner`) is through a `new FileOutputStream(file)`, i.e. without the append-flag, so it should just overwrite. This should at worst affect the first process as the rewrite could corrupt and current reads, but at least process 2 should be fine.
It is not, however, since the symptoms of the issue is that files are permanently corrupted.

So, my simple conclusion is that I'm looking at wrong place. Any ideas? :)

wagon-providers/wagon-file-test/src/main/java/org/apache/maven/wagon/providers/file/LazyLockableFileOutputStreamTestDriver.java
```diff
@@ -0,0 +1,75 @@
...
+                // CHECKSTYLE_ON: MagicNumber
+            }
+            else
+            {
+                throw e;
+            }
+        }
+        System.out.println( "ready" );
+        //noinspection ResultOfMethodCallIgnored
+        System.in.read(); //wait for input to allow test to control when to exit.
```
Tibor17: I do not recommend reading this stream in your code because Surefire receives commands on std/in.

wagon-providers/wagon-file/src/main/java/org/apache/maven/wagon/providers/file/WaitingLockableFileWriter.java
```diff
@@ -0,0 +1,301 @@
...
+     * @param timeoutUnit unit of timeout
+     * @param wait        should we wait while trying to obtain lock
+     * @throws NullPointerException if the file is null
+     * @throws IOException          in case of an I/O error
+     * @since 2.3
+     */
+    public WaitingLockableFileWriter( File file, final Charset encoding, final boolean append, String lockDir,
+                                      long timeout, TimeUnit timeoutUnit, boolean wait ) throws IOException
+    {
+        super();
```
Tibor17: super() is added by compiler, so it can be avoided in code.

Tibor17:
@erikhakansson 
How you know which process is the owner of the lock with the `WaitingLockableFileWriter`?
And what happens if the owner process is killed and the lock file `.lck` presents? Would not be the timeouts on too many locked files hang the wagon too long?

Tibor17:
I guess it would be better if ASF community brings some concept with locking the files.
This activity is too dangerous.
One of concepts can be similar to Gradle which is using the Agent as a single instance of process alive on the machine.

erikhakansson:
@Tibor17 as per discussion with @michael-o, this functionality probably needs to reside elsewhere anyway.

michael-o:
Can we close this for now since it won't land in Wagon anyway? @Tibor17 would you review if @erikhakansson would provide a new PR for Resolver? Your concurrency skills are better than mine.

erikhakansson:
@michael-o sorry for not replying earlier. But yeah we can close this. Unfortunately I probably won't have time anytime soon to look at a new implementation in the Resolver.

Tibor17:
@erikhakansson 
@michael-o 
I have recived the announcement in the email right now. Give me some time to dive into this again.

Tibor17:
@michael-o 
> @Tibor17 would you review if @erikhakansson would provide a new PR for Resolver? Your concurrency skills are better than mine.

No prolem. I will try, feel free to ask me whenever it is necessary.
@erikhakansson 
Maybe one advice. Pls keep it still open since it is important issue for the Maven and we can include more developers. We can continue on development of this PR even if you wouldn't participate.

@michael-o 
@olamy 
Would you mind if we discuss this topic on our mailing list?
We in commercial companies usually write a chart of technical alternatives with Pros/Cons.
This sounds to me like a step which we have not used yet and here it would be a good approach.
WDYT?

michael-o:
@Tibor17 Sure, pick up if you can. I can't, unfortunately.

erikhakansson:
I'll keep the issue open for now then.

If you open up discussions elsewhere it might be good to link them from here so people who end up here can find them.

michael-o:
Superceded by MRESOLVER-131.

PR #50, opened 2018-10-15
WAGON-537 Maven download speed of large artifacts is slow

olaf-otto:
Introduced a blocking buffer strategy that will give filling up the buffer priority to reduce the number of fireTransferProgress invocations.

Introduced a method to dynamically allocate a buffer size better suited for the expected amount of transferred data.

### DIFF
--- a/wagon-provider-api/src/main/java/org/apache/maven/wagon/AbstractWagon.java
+++ b/wagon-provider-api/src/main/java/org/apache/maven/wagon/AbstractWagon.java
@@ -44,6 +44,9 @@ import java.io.InputStream;
 import java.io.OutputStream;
 import java.util.List;
 
+import static java.lang.Math.max;
+import static java.lang.Math.min;
+
 /**
  * Implementation of common facilities for Wagon providers.
  *
@@ -53,6 +56,7 @@ public abstract class AbstractWagon
     implements Wagon
 {
     protected static final int DEFAULT_BUFFER_SIZE = 1024 * 4;
+    protected static final int MAXIMUM_BUFFER_SIZE = 1024 * 512;
 
     protected Repository repository;
 
@@ -560,7 +564,7 @@ public abstract class AbstractWagon
     protected void transfer( Resource resource, InputStream input, OutputStream output, int requestType, long maxSize )
         throws IOException
     {
-        byte[] buffer = new byte[DEFAULT_BUFFER_SIZE];
+        byte[] buffer = bufferForTransferring( resource );
 
         TransferEvent transferEvent = new TransferEvent( this, resource, TransferEvent.TRANSFER_PROGRESS, requestType );
         transferEvent.setTimestamp( System.currentTimeMillis() );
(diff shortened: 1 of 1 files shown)


wagon-provider-api/src/main/java/org/apache/maven/wagon/AbstractWagon.java
```diff
@@ -560,31 +564,78 @@ protected void transfer( Resource resource, InputStream input, OutputStream outp
...
+    protected byte[] bufferForTransferring( Resource resource )
+    {
+        long contentLength = resource.getContentLength();
+
+        if ( contentLength <= 0 )
+        {
+            return new byte[DEFAULT_BUFFER_SIZE];
+        }
+
+        int numberOf4KbSegmentsFor100Chunks = ( ( int ) ( contentLength / ( 4 * 1024 * 100 ) ) );
```
olaf-otto: Kindly asking to specifically review the way the buffer is sized here. 

The Idea is that at least the old buffer size (DEFAULT_BUFFER_SIZE) is used and that buffers should have a size of N * 4Kb. As a rule of thumb, the buffer should be sized such that at least 100 chunks of data are being processed. I've capped the buffer at MAXIMUM_BUFFER_SIZE (512 Kb).

Resulting, most of the files transferred (Metadata such as POMs and checksums, small JARs...) are transferred with the old buffer size. However, larger files are transferred with up to 512 Kb of buffer.

For a 10GB stream, this will reduce the number of chunks from > 5 million (the previous strategy almost never used the entire 4k buffer but less than half) to about 20 thousand chunks.

michael-o: Please use the `static final` for `4 * 1024` and also move the `100` to a `static final`. Also add to the code why you have chosen `100`.

wagon-provider-api/src/main/java/org/apache/maven/wagon/AbstractWagon.java
```diff
@@ -560,31 +564,78 @@ protected void transfer( Resource resource, InputStream input, OutputStream outp
...
+            {
+                nextByte = input.read();
+                if ( nextByte == -1 )
+                {
+                    break;
+                }
+                buffer[read] = ( byte ) nextByte;
+            }
+
+            fireTransferProgress( transferEvent, buffer, read );
```
michael-o: I am somewhat confused why this performs better than described in `#read(byte[], int, int)`'s Javadoc::

> The read(b, off, len) method for class InputStream simply calls the method read() repeatedly ... The default implementation of this method blocks until the requested amount of input data len has been read, end of file is detected, or an exception is thrown. Subclasses are encouraged to provide a more efficient implementation of this method.

I understand that you are mimicking this behavior by filling the buffer manually byte-by-byte, but why is that better?!

michael-o: Makes sense. Have you considered passing to a `BufferInputStream` and see how it scales?

wagon-provider-api/src/main/java/org/apache/maven/wagon/AbstractWagon.java
olaf-otto: Good point, on it!

wagon-provider-api/src/main/java/org/apache/maven/wagon/AbstractWagon.java
michael-o: I'd like to use standard components if possible.

olaf-otto: Hi @michael-o 

Yes, that was actually the first thing I did. However, BufferedInputStream does exactly what was done before in the loop, I.e. try and read *as many bytes as possible without blocking* into a buffer, thus resulting in millions of iterations for large files.

michael-o: Alright, to wrap up: BIS tries to read as many bytes as possible, but you are filling the buffer up to the top to avoid minichunking/fragmentation?

michael-o: Here is the snippet: https://github.com/JetBrains/jdk8u_jdk/blob/master/src/share/classes/sun/security/ssl/AppInputStream.java#L98-L113

So we hit to few `#avaliable()`, right?

olaf-otto: Yes, that's precisely it. I'm filling the buffer to avoid minichunking and indeed it's verly likely we always hit "to few" available() to make use of the entire buffer. Since the latter depends on the underlying network / hardware platform, I'm managing the buffer explicitly, prefering filling the buffer over non-blocking.

olaf-otto: Hi @michael-o ,
Is there anything else that stands in the way of this getting accepted / merged?

jira-importer:
Resolve #591

PR #51, opened 2018-10-25
WAGON-537 Maven transfer speed of large artifacts is slow

olaf-otto:
Implemented a buffer strategy such that filling the buffer to at least 50% has priority over frequent writes.

Added dynamic buffer capacity allocation based on the expected number of bytes to transfer.

Used NIO buffers to simplify buffer management.


olaf-otto:
This change increases download and upload speed of large artifacts by more than a factor of 10 when applied to maven 3.5.4 on Windows 10 64 bit on recent hardware.

jira-importer:
Resolve #591

PR #52, opened 2018-11-11
[MNG-6514] Adding support for encoding basic auth credentials with UT…

aleksgj:
MNG-6514 Saved password with umlaut breaks deployment


jira-importer:
Resolve #596

PR #56, opened 2019-10-03
WAGON-541 Command Line Not Showing ReasonPhrase for Errors

peterlynch:
https://issues.apache.org/jira/browse/WAGON-541

Re-instate status line reason phrases to Wagon HTTP Exception messages
Add consistency to exception messages for all HTTP wagons so that clients ( ie. Maven ) reporting the Exception messages to users have good consistent contextual information. This means also report any custom status line reason phrases that the servers involved might return.


hboutemy:
merged
thank you

jira-importer:
Resolve #604

PR #57, opened 2019-10-09
WAGON-567: support retry on server side errors

zymzxq:
Retry artifact download on server side error

### DIFF
--- a/wagon-providers/wagon-http-shared/src/main/java/org/apache/maven/wagon/shared/http/AbstractHttpClientWagon.java
+++ b/wagon-providers/wagon-http-shared/src/main/java/org/apache/maven/wagon/shared/http/AbstractHttpClientWagon.java
@@ -34,6 +34,7 @@ import org.apache.http.auth.UsernamePasswordCredentials;
 import org.apache.http.client.AuthCache;
 import org.apache.http.client.CredentialsProvider;
 import org.apache.http.client.HttpRequestRetryHandler;
+import org.apache.http.client.ServiceUnavailableRetryStrategy;
 import org.apache.http.client.config.AuthSchemes;
 import org.apache.http.client.config.CookieSpecs;
 import org.apache.http.client.config.RequestConfig;
@@ -61,6 +62,7 @@ import org.apache.http.impl.client.BasicAuthCache;
 import org.apache.http.impl.client.BasicCredentialsProvider;
 import org.apache.http.impl.client.CloseableHttpClient;
 import org.apache.http.impl.client.DefaultHttpRequestRetryHandler;
+import org.apache.http.impl.client.DefaultServiceUnavailableRetryStrategy;
 import org.apache.http.impl.client.HttpClientBuilder;
 import org.apache.http.impl.client.StandardHttpRequestRetryHandler;
 import org.apache.http.impl.conn.PoolingHttpClientConnectionManager;
@@ -451,6 +453,55 @@ public abstract class AbstractHttpClientWagon
         }
     }
 
+    /**
+     * The type of the serviceUnavailableRetryStrategy, defaults to {@code none}.
+     * Values can be {@link default DefaultServiceUnavailableRetryStrategy},
+     * or {@link standard StandardServiceUnavailableRetryStrategy}, or
+     * a fully qualified name class with a no-arg or none to not use a ServiceUnavailableRetryStrategy.
+     */
+    private static final String SERVICE_UNAVAILABLE_RETRY_STRATEGY_CLASS =
+            System.getProperty( "maven.wagon.http.serviceUnavailableRetryStrategy.class", "none" );
+
+    /**
+     * Interval in milliseconds between retries when using a serviceUnavailableRetryStrategy.
+     * <b>1000 by default</b>
+     */
+    private static final int SERVICE_UNAVAILABLE_RETRY_STRATEGY_RETRY_INTERVAL =
+        Integer.getInteger( "maven.wagon.http.serviceUnavailableRetryStrategy.retryInterval", 1000 );
+
+    /**
+     * Maximum number of retries when using a serviceUnavailableRetryStrategy.
+     * <b>5 by default</b>
+     */
+    private static final int SERVICE_UNAVAILABLE_RETRY_STRATEGY_MAX_RETRIES =
+        Integer.getInteger( "maven.wagon.http.serviceUnavailableRetryStrategy.maxRetries", 5 );
+
+    private static ServiceUnavailableRetryStrategy createServiceUnavailableRetryStrategy()
+    {
+        switch ( SERVICE_UNAVAILABLE_RETRY_STRATEGY_CLASS )
+        {
+            case "none": return null;
+            case "default":
+                return new DefaultServiceUnavailableRetryStrategy(
+                    SERVICE_UNAVAILABLE_RETRY_STRATEGY_MAX_RETRIES, SERVICE_UNAVAILABLE_RETRY_STRATEGY_RETRY_INTERVAL );
+            case "standard":
+                return new StandardServiceUnavailableRetryStrategy(
+                    SERVICE_UNAVAILABLE_RETRY_STRATEGY_MAX_RETRIES, SERVICE_UNAVAILABLE_RETRY_STRATEGY_RETRY_INTERVAL );
+            default:
+                try
+                {
+                    final ClassLoader classLoader = AbstractHttpClientWagon.class.getClassLoader();
+                    return ServiceUnavailableRetryStrategy.class.cast(
+                            classLoader.loadClass( SERVICE_UNAVAILABLE_RETRY_STRATEGY_CLASS )
+                                                                          .getConstructor().newInstance() );
+                }
+                catch ( final Exception e )
+                {
+                    throw new IllegalArgumentException( e );
+                }
+        }
+    }
+
     private static Registry<AuthSchemeProvider> createAuthSchemeRegistry()
     {
         return RegistryBuilder.<AuthSchemeProvider>create()
@@ -487,6 +538,7 @@ public abstract class AbstractHttpClientWagon
             .disableConnectionState() //
             .setConnectionManager( httpClientConnectionManager ) //
             .setRetryHandler( createRetryHandler() )
+            .setServiceUnavailableRetryStrategy( createServiceUnavailableRetryStrategy() )
             .setDefaultAuthSchemeRegistry( createAuthSchemeRegistry() )
             .build();
     }
--- a/wagon-providers/wagon-http-shared/src/main/java/org/apache/maven/wagon/shared/http/StandardServiceUnavailableRetryStrategy.java
+++ b/wagon-providers/wagon-http-shared/src/main/java/org/apache/maven/wagon/shared/http/StandardServiceUnavailableRetryStrategy.java
@@ -1,0 +1,79 @@
+package org.apache.maven.wagon.shared.http;
+
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+import org.apache.http.HttpResponse;
+import org.apache.http.HttpStatus;
+import org.apache.http.annotation.Contract;
+import org.apache.http.annotation.ThreadingBehavior;
+import org.apache.http.client.ServiceUnavailableRetryStrategy;
+import org.apache.http.protocol.HttpContext;
+import org.apache.http.util.Args;
+
+/**
+ * An implementation of the {@link ServiceUnavailableRetryStrategy} interface.
+ * that retries {@code 408} (Request Timeout), {@code 429} (Too Many Requests),
+ * and {@code 500} (Server side error) responses for a fixed number of times at a fixed interval.
+ */
+@Contract( threading = ThreadingBehavior.IMMUTABLE )
+public class StandardServiceUnavailableRetryStrategy implements ServiceUnavailableRetryStrategy
+{
+    /**
+     * Maximum number of allowed retries if the server responds with a HTTP code
+     * in our retry code list.
+     */
+    private final int maxRetries;
+
+    /**
+     * Retry interval between subsequent requests, in milliseconds.
+     */
+    private final long retryInterval;
+
+    public StandardServiceUnavailableRetryStrategy( final int maxRetries, final int retryInterval )
+    {
+        super();
+        Args.positive( maxRetries, "Max retries" );
+        Args.positive( retryInterval, "Retry interval" );
+        this.maxRetries = maxRetries;
+        this.retryInterval = retryInterval;
+    }
+
+    @Override
+    public boolean retryRequest( final HttpResponse response, final int executionCount, final HttpContext context )
+    {
+        int statusCode = response.getStatusLine().getStatusCode();
+        boolean retryableStatusCode = statusCode == HttpStatus.SC_REQUEST_TIMEOUT
+                // Too Many Requests ("standard" rate-limiting)
+                || statusCode == AbstractHttpClientWagon.SC_TOO_MANY_REQUESTS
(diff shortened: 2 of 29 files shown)


wagon-providers/wagon-http-shared/src/main/java/org/apache/maven/wagon/shared/http/AbstractHttpClientWagon.java
```diff
@@ -487,6 +487,8 @@ private static CloseableHttpClient createClient()
             .disableConnectionState() //
             .setConnectionManager( httpClientConnectionManager ) //
             .setRetryHandler( createRetryHandler() )
+            .setServiceUnavailableRetryStrategy(
+                    new SimpleServiceUnavailableRetryStrategy( 10, 1000 ) )
```
olamy: if we implement this it should be optional and retry number/time between retry configurable.

michael-o:
IS this one different from the `createRetryHandler()`?

ColinHebert:
@michael-o `retryHandler` covers connection errors whereas `serviceUnavailableRetryStrategy` covers service errors (HTTP error codes).

wagon-providers/wagon-http-shared/src/main/java/org/apache/maven/wagon/shared/http/AbstractHttpClientWagon.java
```diff
@@ -451,6 +453,54 @@ private static HttpRequestRetryHandler createRetryHandler()
...
+     */
+    private static final String SERVICE_UNAVAILABLE_RETRY_STRATEGY_CLASS =
+            System.getProperty( "maven.wagon.http.serviceUnavailableRetryStrategy.class", "null" );
+
+    /**
+     * Interval in milliseconds between retries when using a serviceUnavailableRetryStrategy.
+     * <b>1000 by default</b>
+     */
+    private static final int SERVICE_UNAVAILABLE_RETRY_STRATEGY_RETRY_INTERVAL =
+        Integer.parseInt( System.getProperty( "maven.wagon.http.serviceUnavailableRetryStrategy.retryInterval", "1000" ) );
```
michael-o: Use `Integer.getIntger()` for that.

wagon-providers/wagon-http-shared/src/main/java/org/apache/maven/wagon/shared/http/AbstractHttpClientWagon.java
```diff
@@ -451,6 +453,54 @@ private static HttpRequestRetryHandler createRetryHandler()
...
     }
 
+    /**
+     * The type of the serviceUnavailableRetryStrategy, defaults to {@code null}.
+     * Values can be {@link default DefaultServiceUnavailableRetryStrategy},
+     * or {@link standard SimpleHttpRequestRetryHandler},
+     * a fully qualified name class with a no-arg or null to not use a ServiceUnavailableRetryStrategy.
+     */
+    private static final String SERVICE_UNAVAILABLE_RETRY_STRATEGY_CLASS =
+            System.getProperty( "maven.wagon.http.serviceUnavailableRetryStrategy.class", "null" );
```
michael-o: Why not use a real `null`? Or better `none`.

wagon-providers/wagon-http-shared/src/main/java/org/apache/maven/wagon/shared/http/AbstractHttpClientWagon.java
```diff
@@ -451,6 +453,54 @@ private static HttpRequestRetryHandler createRetryHandler()
...
+     * <b>1000 by default</b>
+     */
+    private static final int SERVICE_UNAVAILABLE_RETRY_STRATEGY_RETRY_INTERVAL =
+        Integer.parseInt( System.getProperty( "maven.wagon.http.serviceUnavailableRetryStrategy.retryInterval", "1000" ) );
+
+    /**
+     * Maximum number of retries when using a serviceUnavailableRetryStrategy.
+     * <b>5 by default</b>
+     */
+    private static final int SERVICE_UNAVAILABLE_RETRY_STRATEGY_MAX_RETRIES =
```
michael-o: Use `Integer.getIntger()` for that.

wagon-providers/wagon-http-shared/src/main/java/org/apache/maven/wagon/shared/http/StandardServiceUnavailableRetryStrategy.java
```diff
@@ -28,17 +28,13 @@
...
 
 /**
- * An implementation of the {@link SimpleServiceUnavailableRetryStrategy} interface.
+ * An implementation of the {@link ServiceUnavailableRetryStrategy} interface.
  * that retries {@code 408} (Request Timeout), {@code 420} (Method Failure), {@code 429} (Too Many Requests),
  * and {@code 500} (Server side error) responses for a fixed number of times at a fixed interval.
  */
 @Contract( threading = ThreadingBehavior.IMMUTABLE )
-public class SimpleServiceUnavailableRetryStrategy implements ServiceUnavailableRetryStrategy
+public class StandardServiceUnavailableRetryStrategy implements ServiceUnavailableRetryStrategy
```
michael-o: I am not convinced that this should be really called standard because of >= 500 and there seems to many intepretations of 420 and all of them are deprecated or not present: https://www.whoishostingthis.com/resources/http-status-codes/#method-failure-400.

I am fine with 408, 421, 503 and 504. 

Please tell me why all 5xy are necessary.

zymzxq:
Agree >=500 is too coarse grained. I'd also like to check 500 and 502 as well. AWS s3, which we redirect our maven download requests to,  returns 500 when it has problems and expects client to retry. We also experienced s3 returned 502 when it had issues. AWS document: https://stash.atlassian.com/projects/BUILDENG/repos/buildeng-plan-templates/pull-requests/1024/overview

Happy to remove the non standard 420 check.

wagon-providers/wagon-http-shared/src/main/java/org/apache/maven/wagon/shared/http/AbstractHttpClientWagon.java
```diff
@@ -451,6 +453,55 @@ private static HttpRequestRetryHandler createRetryHandler()
         }
     }
 
+    /**
+     * The type of the serviceUnavailableRetryStrategy, defaults to {@code null}.
```
michael-o: Please update.

wagon-providers/wagon-http-shared/src/main/java/org/apache/maven/wagon/shared/http/AbstractHttpClientWagon.java
```diff
@@ -451,6 +453,55 @@ private static HttpRequestRetryHandler createRetryHandler()
...
+     * Values can be {@link default DefaultServiceUnavailableRetryStrategy},
+     * or {@link standard StandardServiceUnavailableRetryStrategy},
```
michael-o: put the "or" after that.

wagon-providers/wagon-http-shared/src/main/java/org/apache/maven/wagon/shared/http/AbstractHttpClientWagon.java
```diff
@@ -451,6 +453,55 @@ private static HttpRequestRetryHandler createRetryHandler()
...
+     * a fully qualified name class with a no-arg or null to not use a ServiceUnavailableRetryStrategy.
```
michael-o: null > none

wagon-providers/wagon-http/src/test/java/org/apache/maven/wagon/providers/http/AbstractHttpClientWagonTest.java
```diff
@@ -58,7 +58,7 @@
 {
     @Ignore("This test is validating nothing and require internet connection which we should avoid so ignore it")
     public void test()
-        throws Exception
+            throws Exception
```
michael-o: Please drop formattting changes in this class, they are unrelated

michael-o:
Way better now, I will try to have a look at this this week.

michael-o:
I will need a couple of days more, but this defitinitively worth merging. I had a similar requirement when Nexus requires too long to download a resources HTTPd would drop the request with 408.

wagon-providers/wagon-http-shared/src/main/java/org/apache/maven/wagon/shared/http/StandardServiceUnavailableRetryStrategy.java
```diff
@@ -0,0 +1,80 @@
...
+import org.apache.http.HttpStatus;
+import org.apache.http.annotation.Contract;
+import org.apache.http.annotation.ThreadingBehavior;
+import org.apache.http.client.ServiceUnavailableRetryStrategy;
+import org.apache.http.protocol.HttpContext;
+import org.apache.http.util.Args;
+
+/**
+ * An implementation of the {@link ServiceUnavailableRetryStrategy} interface.
+ * that retries {@code 408} (Request Timeout), {@code 420} (Method Failure), {@code 429} (Too Many Requests),
```
michael-o: 420 is gone.

ColinHebert:
@michael-o Currently the default behaviour is to keep the `ServiceUnavailableRetryStrategy` unset by default (`none`). And the current backoff mechanism is still enabled by default, so there should not be an issue with rolling that out (consider `ServiceUnavailableRetryStrategy` as an experimental flag that is being released to be tested).

If we're satisfied with the retryStrategy flag as a replacement of the backoff mechanism then we should start doing the following:
 - Set the **default** backoff settings to 0 _if the retryStrategy is set_ (effectively disabling backoff)
 - Warn if **both** backoff and retryStrategy are explicitly configured

If we're still okay with the retryStrategy being the way forward, we should:
 - Set a default retryStrategy that for backward compatibility handles 429s (`standard` might be it, or maybe a new one that handles only 429s), and unconditionally set the **default** backoff settings to 0
 - Warn if the backoff strategy is explicitly set and recommend moving to a custom retry strategy if needed.

Then after a certain period of time:
 - Remove the backoff mechanism entirely

Some of those steps can be merged together for a faster rollout/change but the gist of it is, by default the retryStrategy is not incompatible with the backoff mechanism (it is redundant and might try to a few more retries in case of a 429 specifically though, which shouldn't be a huge problem).

Thoughts?

zymzxq:
I like the progressive approach that Colin suggested.

michael-o:
@ColinHebert I really much like your proposal. As soon as this goes in I will deprecate and mark the current backoff mechanism for removal in 4.0.0.

I am considering to modify `DefaultServiceUnavailableRetryStrategy` by accepting a set of possible status codes. I will discuss this with other HttpComponents committers since I am one.

I think that the current backoff mechanism will be unavailable as soon as a new handler pops in or it will fail and the mech would pop in.

My proposal is:

* Lower default retry to 3 because the low level retry handler uses 3 too
* Raise from default 1000 ms to 2000 ms to 2500 ms not to bomb a server and have a closer value to `maven.wagon.httpconnectionManager.backoffSeconds`
* Introduce a Maven-default ("default") which handles 503 and 429.
* Rename the custom impl to "extended" because it covers way more status codes.

WDY?

ColinHebert:
Makes sense for `DefaultServiceUnavailableRetryStrategy`, I'm actually surprised 429 was not handled out of the box.

Regarding the proposal:
 - Makes sense for the default retry to be more reasonable. In CI/CD context the retry might need to be higher but for local development it's perfectly acceptable
 - I am not sure you are going to be able to raise that time. The entire retry mechanism is encapsulated AFAIK into the HTTP client. What should really be happening there is having a backoff mechanism in the HTTP Client itself, there are BackoffManager as well as ConnectionBackoffStrategy that can be set, but I'm not familiar with their behaviour (I only know that they're not set by default)
 - Makes sense again, it would be interesting to compare to prior art in that domain. I believe that @zymzxq started with something similar to npm https://github.com/npm/make-fetch-happen/blob/latest/index.js#L386-L392. Another interesting one is how gradle handles it: https://github.com/gradle/gradle/blob/master/subprojects/dependency-management/src/main/java/org/gradle/api/internal/artifacts/repositories/transport/NetworkingIssueVerifier.java#L51
 - Ultimately this naming will have very little impact on the end user, so I'm indifferent to the naming here.

michael-o:
I will inquire why 429 is not included in the default config.

* Alright, shall do so
* I haven't even considered both Backoff interfaces. I don't even know whether a backoff is suited for 503 or rather 429, I guess for 429 only (throttling). I'd like to keep complexity low here and don't open new wormholes. Do you think that using the `ServiceUnavailableRetryStrategy` is really the wrong way to go?
* Both NPM and Gradle approaches seem to generic for me, saying all 5xy can be repeated does not make sense. I would logically limit it to 503, 504 and at most 502. Everything else won't be solved by a retry.

ColinHebert:
> Do you think that using the ServiceUnavailableRetryStrategy is really the wrong way to go?

I think it's "good enough" to begin with. This is a strict upgrade to the current behaviour which is failing fast on behaviour that we know we should retry. Keeping complexity low could be either dropping backoff entirely (if we're retrying 3 times only, a backoff strategy feels unnecessary) or delegating that entirely to the HTTPClient (but as I said I'm not well versed into HTTPClient's ability to backoff out of the box).

> Both NPM and Gradle approaches seem to generic for me, saying all 5xy can be repeated does not make sense.

I believe that the intent is more along the lines of "we do not expect a 5xx from a user-configured repository on a fairly simple GET request, therefore we might want to consider it a transient error". If you wish to limit it though, I would strongly suggest to add 500 to the list of handled errors as many HTTP Servers use it as a catchall (therefore it has a non zero probably of being simply retryable)

michael-o:
@ColinHebert  OK, I guess I need to discuss with other HttpComponents devs whether we should really split client and servers errors. `BackoffManager` for client errors and the `ServiceUnavailable...` stuff for the server side. If I won't get a reaction from others, I will go the route we have discussed here.

@ok2c Can you drop a few words here? Thanks!

ok2c:
@michael-o @ColinHebert  The original intent of `ConnectionBackoffStrategy` and `BackoffManager` was to dynamically reduce connection pool limits in case of the server overload. It should have no bearing on automatic request re-execution or recovery. Protocol level recovery logic should be handled by `ServiceUnavailableRetryStrategy` and transport level recovery logic should be handled by `HttpRequestRetryHandler`. 

I am not even sure if WAGON needs `ConnectionBackoffStrategy` and `BackoffManager` at all.

Does this help in any way?

ColinHebert:
So if I understand @ok2c, there is no out-of-the-box way to do backoff on protocol level retries, as `ServiceUnavailableRetryStrategy` provides a retry interval which does not carry information on how many previous attempts have been made?

The only thing that could be reasonably done for a given request is adding jitter to the retry interval (which really isn't necessary).

Is that correct?


Also regarding the backoff strategies in wagon, that makes sense to me and lines up with my previous assumption "if we're retrying 3 times only by default, a backoff strategy feels unnecessarily complex to begin with".

ok2c:
> So if I understand @ok2c, there is no out-of-the-box way to do backoff on protocol level retries, as ServiceUnavailableRetryStrategy provides a retry interval which does not carry information on how many previous attempts have been made?

@ColinHebert This is correct. There are two options here: 
* in 5.0 we can still extend or rework `ServiceUnavailableRetryStrategy` but one needs to hurry. 
* In 4.5.x one can replace `ServiceUnavailableRetryExec` with a custom request execution handler.

ColinHebert:
For 5.0 it seems trivial to add `c` as a parameter to `getRetryInterval()` (similar to `retryRequest()`).
https://github.com/apache/httpcomponents-client/blob/07a9714f3b89509b231fddbb667eb79073e9872d/httpclient5/src/main/java/org/apache/hc/client5/http/impl/classic/ServiceUnavailableRetryExec.java#L94

For 4.5.x and Wagon, I suppose that it leaves us with the original option which is deprecating backoff for the time being (and eventually reintroduce it when moving to HttpClient 5.x?), instead use the native support for protocol level retries.

ok2c:
> For 5.0 it seems trivial to add c as a parameter to getRetryInterval() (similar to retryRequest())

@ColinHebert I will happily review a PR for this change.

michael-o:
@ok2c thank you for the evaluation.

@ColinHebert Can you provide that PR? Would that be the backoff you want to introduce with the usage of HttpClient 5.0?

I will go on the current approach with the `ServiceUnavailableRetryStrategy` for both client and server errors.

zymzxq:
Note: in httpclient5 [DefaultServiceUnavailableRetryStrategy](https://github.com/apache/httpcomponents-client/blob/master/httpclient5/src/main/java/org/apache/hc/client5/http/impl/DefaultServiceUnavailableRetryStrategy.java#L84) respects retryAfter response header before falling back to default interval.

michael-o:
@ok2c , I think this PR (for the `c`) won't be anymore necessary because of HTTPCLIENT-2019.

michael-o:
An important note, this will be sooner or later be merged into a new `HttpRequestRetryStrategy` impl in future version when we switch to HC 5.0.

michael-o:
@ColinHebert Please have a look at https://github.com/apache/httpcomponents-client/pull/178.

jira-importer:
Resolve #618

PR #58, opened 2019-10-27
[WAGON-569] Inconsistent encoding behavior for repository URLs with s…

michael-o:
…paces

@ok2c Can you have a look? I hope I have properly used the `URLEncodedUtils`.

### DIFF
--- a/wagon-providers/wagon-http-shared/src/main/java/org/apache/maven/wagon/shared/http/EncodingUtil.java
+++ b/wagon-providers/wagon-http-shared/src/main/java/org/apache/maven/wagon/shared/http/EncodingUtil.java
@@ -24,6 +24,8 @@ import java.net.URI;
 import java.net.URISyntaxException;
 import java.net.URL;
 
+import org.apache.http.client.utils.URLEncodedUtils;
+
 /**
  * Encoding utility.
  *
@@ -38,7 +40,9 @@ public class EncodingUtil
      * @return Parsed/encoded {@link URI} that represents the string form URL passed in.
      * @throws MalformedURLException
      * @throws URISyntaxException
+     * @deprecated to be removed with 4.0.0
      */
+    @Deprecated
     public static URI encodeURL( String url )
         throws MalformedURLException, URISyntaxException
     {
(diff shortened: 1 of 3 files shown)


wagon-providers/wagon-http-shared/src/main/java/org/apache/maven/wagon/shared/http/EncodingUtil.java
```diff
@@ -77,29 +83,28 @@ public static String encodeURLToString( String url )
...
      * @since TODO
      */
     public static String encodeURLToString( String baseUrl, String... paths )
     {
         StringBuilder url = new StringBuilder( baseUrl );
 
-        String[] parts = paths == null ? //
+        String[] segments = paths == null ? //
             new String[0] : //
             paths.length == 1 ? paths[0].split( "/" ) : paths;
```
ok2c: @michael-o This part looks a bit dodgy to me. I would rather use two distinct methods here `encodeURLToString( String baseUrl, String... pathSegments )` and `encodeURLToString( String baseUrl, String path )` with the latter splitting the path into segments and then calling the former. Otherwise looks OK to me.

michael-o: This might be problematic because how should the compiler know `#encodingURLToString("../dd", "path/d/c")` should not be passed to the varargs overload?

ok2c: @michael-o Different (more descriptive) method names?

michael-o: I fiddled with it a bit on Zulu 7, Oracle JDK 7 + 11 and the overload is properly inferred. There is only one edge case. When you have the path *segment* `a/b` and you pass it, the single string overload will be used. You have to wrap into an array.

michael-o: As soon as Jenkins build passes, I will merge into master. Thank you again for the advise.

PR #59, opened 2019-10-30
[WAGON-573] EntityUtils.consumeQuietly() never called on non-2xx stat…

michael-o:
…us codes

@ok2c Please have a look whether this makes sense to properly clean up resources.


michael-o:
@ok2c, thanks for the review. Went through the code and and the HttpClient tutorial. I do not fully understand why the tutorial does both, consuming the response entiry and closing the response. So does the Wagon code. Should one do either one depending on the status code and problem?

jira-importer:
Resolve #641

PR #60, opened 2019-11-08
[WAGON-574] advertise jsoup license for shaded jar

hboutemy:
https://issues.apache.org/jira/browse/WAGON-574

### DIFF
--- a/wagon-providers/wagon-http/pom.xml
+++ b/wagon-providers/wagon-http/pom.xml
@@ -33,6 +33,19 @@ under the License.
     Wagon provider that gets and puts artifacts through HTTP(S) using Apache HttpClient-4.x.
   </description>
 
+  <licenses>
+    <license>
+      <name>Apache License, Version 2.0</name>
+      <url>https://www.apache.org/licenses/LICENSE-2.0.txt</url>
+      <distribution>repo</distribution>
+    </license>
+    <license>
+      <name>MIT License</name>
+      <url>https://raw.githubusercontent.com/jhy/jsoup/jsoup-1.12.1/LICENSE</url>
+      <distribution>repo</distribution>
+      <comments>shaded jar is bundled with jsoup, Copyright (c) 2009-2019 Jonathan Hedley (jonathan@hedley.net)</comments>
+    </license>
+  </licenses>
 
   <dependencies>
     <dependency>


wagon-providers/wagon-http/pom.xml
```diff
@@ -33,6 +33,19 @@ under the License.
     Wagon provider that gets and puts artifacts through HTTP(S) using Apache HttpClient-4.x.
   </description>
 
+  <licenses>
+    <license>
+      <name>Apache License, Version 2.0</name>
+      <url>https://www.apache.org/licenses/LICENSE-2.0.txt</url>
+      <distribution>repo</distribution>
+    </license>
+    <license>
```
eolivelli: I am not sure about the official meaning of this tag.
Usually crawlers read this tag to classify the project.

This project is not MIT Licensed and it is not dual licensed.

Maybe it is better to add the 'comment' into the primary ALv2.0 license tag

michael-o:
I'd rather drop JSoup usage because of the reasons I have described on the mailing list.

eolivelli:
@michael-o

I missed your mail.
Dropping jsoup will be very nice!

wagon-providers/wagon-http/pom.xml
hboutemy: Maven defined the format, with following documentation https://maven.apache.org/ref/3.6.2/maven-model/maven.html#class_license

if you want to add a comment, just propose a content

hboutemy:
how can you be affirmative that removing jsoup won't break any case?
I don't want to have a Maven 3.6.4 release because of some edge case
the feature was added in WAGON-58 https://issues.apache.org/jira/browse/WAGON-58

hboutemy:
> This project is not MIT Licensed
the shaded jar has many licences

wagon-providers/wagon-http/pom.xml
```diff
@@ -33,6 +33,19 @@ under the License.
     Wagon provider that gets and puts artifacts through HTTP(S) using Apache HttpClient-4.x.
   </description>
 
+  <licenses>
+    <license>
+      <name>Apache License, Version 2.0</name>
+      <url>https://www.apache.org/licenses/LICENSE-2.0.txt</url>
+      <distribution>repo</distribution>
```
eolivelli: Here I would add:
<comments>shaded jar is bundled with jsoup that it MIT licensed, Copyright (c) 2009-2019 Jonathan Hedley (jonathan@hedley.net), see https://raw.githubusercontent.com/jhy/jsoup/jsoup-1.12.1/LICENSE</comments>

eolivelli:
> I don't want to have a Maven 3.6.4 release because of some edge case.

fair

michael-o:
@hboutemy First of all, Maven Core does not rely on that. Can you search for method in our code case?! Second, HTTP does not provide any listing of files. The implementation is highly coupled to a 10+ y/o Apache HTTPd mod_userdir output. It was a misconception from the beginning.

hboutemy:
@michael-o 

> First of all, Maven Core does not rely on that. Can you search for method in our code case?!

sure, core does not, but core provides Wagon implementation to plugins: the issue is more subtle, people will have features that may not work with newer Maven versions

> The implementation is highly coupled to a 10+ y/o Apache HTTPd mod_userdir output.

you mean that it does not work for current central directory listing, like https://repo.maven.apache.org/maven2/org/apache/maven/wagon/ ? or it does not work for classical repository manager directory listing, like https://repository.apache.org/content/groups/snapshots/org/apache/maven/wagon/wagon/ ?

> It was a misconception from the beginning.

it's usual for repository managers to provide directory listing, it's a useful idea

Now:
1. once we figure out if the feature is really working or not, we could perhaps avoid this jsoup dependency: html we're looking for is probably not that complex
2. the MIT license in artifact's pom.xml with comment looks to me a reasonable quick first step

michael-o:
@hboutemy, I did not test anything. The interface implies that the underlying protocol has a notion of directories and collections. HTTP has neither. Assuming that some server will deliver some collection is wrong because it breaks a basic principle: never do an assumption about an unknown environment. Having that said, the given implementation cannot work reliably on any type of HTML style listing. Also Wagon is not necessary limited for repository manager.

All I am saying is that if a protocol does not define collections/directories, don't try to implement them by some crude means with some assumptions, simply return `UnimplementedException`. Since HTTP really does not, see RFC 7230+, deprecate and mark it for removal.

eolivelli:
Can't we keep jsoup as simple dependency and hide it somehow in maven core with some plexus classloader magic?

Just thinking out loud...
If @michael-o assumption is correct we could simply not bundle it on maven-core at all, if someone hits a ClassNotFoundException he can add the jsoup jar to the lib directory.

michael-o:
> Just thinking out loud...
> If @michael-o assumption is correct we could simply not bundle it on maven-core at all, if someone hits a ClassNotFoundException he can add the jsoup jar to the lib directory.

I really like this idea!

michael-o:
I just grepped through code and resolver and `filelist` (ci) is not called.

slachiewicz:
Why not simple Rexexp like `<a[^>]* href="([^"]*)"` ?

michael-o:
Because if two reasons:

* regex and HTML/XML don't play well
* there is no guarantee that the listing will be in the format you assume

slachiewicz:
Ok, 1st point sounds reasonable. For the 2nd I think above we have arguments to not have listing functionality..

So maybe this solution https://github.com/apache/maven-wagon/commit/490362ba31dc5d8e119ffd447390327bf9d988a0 ?

michael-o:
Attention, this might require a JDK with X11 support because it is a Swing component. I'd rather throw unsupported operation exception. You can't implement something which is not specified.

olamy:
@slachiewicz please not using `javax.swing` :)

slachiewicz:
What to say.. https://github.com/apache/maven-doxia/blob/04070b889a2489d9b976b5fc865fd4e4e863b5c8/doxia-modules/doxia-module-xhtml5/src/main/java/org/apache/maven/doxia/module/xhtml5/Xhtml5Sink.java .. :-)))

michael-o:
> 
> 
> What to say.. https://github.com/apache/maven-doxia/blob/04070b889a2489d9b976b5fc865fd4e4e863b5c8/doxia-modules/doxia-module-xhtml5/src/main/java/org/apache/maven/doxia/module/xhtml5/Xhtml5Sink.java .. :-)))

True case, that code is really old.

michael-o:
JSoup has been removed, time to close this.

jira-importer:
Resolve #621

PR #61, opened 2019-11-25
Fix cast exception, when the server returns an xml containing a newli…

zawn:
Fix cast exception, when the server returns an xml containing a newline, property.getValue() is an ArrayList

eg。
<?xml version="1.0"?>
<d:multistatus xmlns:d="DAV:" xmlns:s="http://sabredav.org/ns" xmlns:oc="http://owncloud.org/ns">
 <d:response>
  <d:href>/remote.php/webdav/</d:href>
  <d:propstat>
   <d:prop>
    <d:resourcetype>
     <d:collection/>
    </d:resourcetype>
   </d:prop>
   <d:status>HTTP/1.1 200 OK</d:status>
  </d:propstat>
 </d:response>
</d:multistatus>


zawn:
image

michael-o:
Please create the JIRA issue with the failure first.

zawn:
Sorry, I learned a bit, the test case uses Jetty, but I don't know how to use Jetty to simulate the above response

zawn:
@eolivelli

michael-o:
I do not even understand the problem here. Can you at least elaborate?

michael-o:
No reaction for months.

PR #62, opened 2020-02-22
[WAGON-570] Use RedirectStrategy from HttpClient rather than our cust…

michael-o:
…om code

@ok2c Please have a look here.

### DIFF
--- a/wagon-providers/wagon-http-shared/src/main/java/org/apache/maven/wagon/shared/http/WagonRedirectStrategy.java
+++ b/wagon-providers/wagon-http-shared/src/main/java/org/apache/maven/wagon/shared/http/WagonRedirectStrategy.java
@@ -1,0 +1,87 @@
+package org.apache.maven.wagon.shared.http;
+
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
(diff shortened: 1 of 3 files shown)


wagon-providers/wagon-http-shared/src/main/java/org/apache/maven/wagon/shared/http/WagonRedirectStrategy.java
```diff
@@ -0,0 +1,99 @@
...
+        {
+            return new HttpHead( uri );
+        }
+        else if ( method.equalsIgnoreCase( HttpGet.METHOD_NAME ) )
+        {
+            return new HttpGet( uri );
+        }
+        else
+        {
+            return RequestBuilder.copy( request ).setUri( uri ).build();
```
ok2c: @michael-o I suppose there is no need to treat `HEAD` and `GET` as special cases. `RequestBuilder.copy( request ).setUri( uri ).build();` should be OK for all methods. 
Looks good to me.

michael-o: That's actually from here:  https://github.com/apache/httpcomponents-client/blob/4.5.x/httpclient/src/main/java/org/apache/http/impl/client/DefaultRedirectStrategy.java#L222-L231

ok2c: @michael-o I know. There however it actually makes sense. By default HttpClient 4.5 makes a full copy of the redirected request only in response to status 307.

jira-importer:
Resolve #636

PR #64, opened 2020-03-29
WAGON-585 maven.wagon.http.ssl features documentation is incomplete

Alexander-Kiselyov:
More details regarding how Wagon's behavior changes with setting particular properties to true added. Dependency on an insecure property described.

### DIFF
--- a/wagon-providers/wagon-http/src/site/apt/index.apt
+++ b/wagon-providers/wagon-http/src/site/apt/index.apt
@@ -50,11 +50,11 @@ Features
 
  Other features can be configured through system properties:
 
- * <<<maven.wagon.http.ssl.insecure>>> = true/false (default false), enable/disable relaxed check of public key certificates (e.g. self-signed ones). Relaxed check means that any chain with 1 or more certificates will be considered valid if all the certificate dates in the chain are valid (dates check can be overridden as well - see below). Setting it to <<<true>>> also enables usage of the following properties:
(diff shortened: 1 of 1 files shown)


wagon-providers/wagon-http/src/site/apt/index.apt
```diff
@@ -50,11 +50,11 @@ Features
 
  Other features can be configured through system properties:
 
- * <<<maven.wagon.http.ssl.insecure>>> = true/false (default false), enable/disable use of relaxed ssl check for user generated certificates.
+ * <<<maven.wagon.http.ssl.insecure>>> = true/false (default false), enable/disable relaxed check of public key certificates (e.g. self-signed ones). Relaxed check means that any chain with 1 or more certificates will be considered valid if all the certificate dates in the chain are valid (dates check can be overridden as well - see below). Setting it to <<<true>>> also enables usage of the following properties:
 
- * <<<maven.wagon.http.ssl.allowall>>> = true/false (default false), enable/disable match of the server's X.509 certificate with hostname. If disabled, a browser like check will be used.
+    * <<<maven.wagon.http.ssl.allowall>>> = true/false (default false), enable/disable match of the server's X.509 certificate with hostname. If <<<false>>>/unset, a "browser-like check" will be used, which means that the hostname must match either the first CN, or any of the subject-alts. A wildcard can occur in the CN, and in any of the subject-alts.
```
michael-o: Browsers are moving fast and that has nothing to do with browsers, but is a RFC 5280 requirement. Let's throw "browsers" out.

michael-o: rename `subject-alts` either to `SAN` or the full version.

Alexander-Kiselyov: Sure. Frankly, I just copy-pasted the javadoc of `org.apache.http.conn.ssl.BrowserCompatHostnameVerifier` that will be used in this case.

Alexander-Kiselyov: Will do, was just trying to be less intrusive.

PR #66, opened 2020-05-11
[SCM-930] update plexus-utils

elharo:
@michael-o This syncs the version with Maven core head which already has 3.3.0


elharo:
Jenkins passes.

PR #67, opened 2020-05-18
Feature/MNG-5583 per endpoint support for PKI authentication

spyhunter99:


spyhunter99:
hang on,forgot i had disabled checkstyle.  sorry stand by

wagon-provider-api/src/main/java/org/apache/maven/wagon/authentication/AuthenticationInfo.java
```diff
@@ -51,7 +52,300 @@
      * The absolute path to private key file
      */
     private String privateKey;
+    
+     /**
+     *
```
mthmulders, line 57: Nit-picking: I don't like the superfluous `*` lines. Also, I think the Javadoc specs say they shouldn't be there.

wagon-provider-api/src/main/java/org/apache/maven/wagon/authentication/AuthenticationInfo.java
```diff
@@ -51,7 +52,300 @@
...
+    private String trustStore;
+
+    /**
+     *
+     *
+     * The password to the trust store.
+     *
+     *
+     */
+    private String trustStorePassword;
```
mthmulders, line 73: This field contains a password. We must convert it to `char[]` later on, so can't we store it as a `char[]`?

wagon-provider-api/src/main/java/org/apache/maven/wagon/authentication/AuthenticationInfo.java
```diff
@@ -51,7 +52,300 @@
...
+    private String keyStore;
+
+    /**
+     *
+     *
+     * Keystore password, can be null
+     *
+     * .
+     */
+    private String keyStorePassword;
```
mthmulders, line 100: This field contains a password. We must convert it to `char[]` later on, so can't we store it as a `char[]`?

wagon-provider-api/src/main/java/org/apache/maven/wagon/authentication/AuthenticationInfo.java
```diff
@@ -51,7 +52,300 @@
...
+    private String keyAlias;
+
+    /**
+     *
+     *
+     * The password to unlock the key, can be null
+     *
+     * .
+     */
+    private String keyPassword;
```
mthmulders, line 120: This field contains a password. We must convert it to `char[]` later on, so can't we store it as a `char[]`?

wagon-providers/wagon-http-shared/src/main/java/org/apache/maven/wagon/shared/http/AbstractHttpClientWagon.java
```diff
@@ -593,13 +603,189 @@ private static CloseableHttpClient createClient()
...
+                if ( file.exists() ) 
+                {
+                    fis = new FileInputStream( file );
+                } 
+                //also a null password is fine with windows and even with JKS files
+                char[] keyPass = authenticationInfo.getTrustStorePassword() != null
+                        ? authenticationInfo.getTrustStorePassword().toCharArray()
+                        : null;
+                keystore.load( fis, keyPass );
+                if ( keyPass != null )
```
mthmulders, line 642: I assume you're doing this to clean up the password from memory, but it currently still resides in the `authenticationInfo` variable. So I think it has little to no effect.

wagon-providers/wagon-http-shared/src/main/java/org/apache/maven/wagon/shared/http/AbstractHttpClientWagon.java
```diff
@@ -593,13 +603,189 @@ private static CloseableHttpClient createClient()
...
+                }
+                String alg = KeyManagerFactory.getDefaultAlgorithm();
+                TrustManagerFactory fac = TrustManagerFactory.getInstance( alg );
+                fac.init( keystore );
+                trustManagers = fac.getTrustManagers();
+            }
+
+            if ( authenticationInfo.getKeyStore() != null ) 
+            {
+                KeyStore keystore = KeyStore.getInstance( authenticationInfo.getKeyStoreType() == null ? "JKS"
```
mthmulders, line 661: As far as I can see, the code for loading the Keystore is identical to lines 627 - 652 (roughly). Can't we refactor this into a method?

wagon-providers/wagon-http-shared/src/main/java/org/apache/maven/wagon/shared/http/AbstractHttpClientWagon.java
```diff
@@ -593,13 +603,189 @@ private static CloseableHttpClient createClient()
...
+                    //let's make sure it exists in the key store that it has a 
+                    //key pair
+                    if ( keystore.containsAlias( alias ) ) 
+                    {
+                        if ( !keystore.isKeyEntry( alias ) ) 
+                        {
+                            alias = null;
+                        }
+                    } 
+                    else if ( keystore.containsAlias( alias.toLowerCase() ) )
```
mthmulders, line 703: Why this additional check on the lower-cased variant?

wagon-providers/wagon-http-shared/src/main/java/org/apache/maven/wagon/shared/http/AbstractHttpClientWagon.java
```diff
@@ -593,13 +603,189 @@ private static CloseableHttpClient createClient()
...
+        try 
+        {
+            if ( authenticationInfo.getTrustStore() != null ) 
+            {
+                KeyStore keystore = KeyStore.getInstance( authenticationInfo.getTrustStoreType() == null ? "JKS"
+                        : authenticationInfo.getTrustStoreType() );
+                FileInputStream fis = null;
+                //on windows platforms, the truststoreType of "WINDOWS" mounts
+                //to the windows certificate store, so a null input stream is just fine
+                File file = new File( ( authenticationInfo.getTrustStore() ) );
```
mthmulders, line 632: Using Java 7's `try-with-resources` might make the code from line 632 - 652 a little more compact.

wagon-provider-api/src/main/java/org/apache/maven/wagon/authentication/AuthenticationInfo.java
```diff
@@ -24,9 +24,10 @@
 /**
  * This class holds the set of properties used when instance of the <code>Wagon</code>
  * will use during login operation.
+ * <br>May 2020, added PKI settings, see MNG-5583
```
mthmulders, line 27: I'm not sure we need a changelog in the Javadoc.

wagon-provider-api/src/main/java/org/apache/maven/wagon/AbstractWagon.java
```diff
@@ -36,21 +49,6 @@
 import org.apache.maven.wagon.resource.Resource;
 import org.codehaus.plexus.util.IOUtil;
 
-import java.io.File;
```
michael-o, line 39: Skip the reformat, unrelated.

wagon-provider-api/src/main/java/org/apache/maven/wagon/authentication/AuthenticationInfo.java
michael-o, line 27: We have version control. This should be at most in the commit message.

wagon-provider-api/src/main/java/org/apache/maven/wagon/authentication/AuthenticationInfo.java
```diff
@@ -24,9 +24,10 @@
...
  *
  * @author <a href="michal.maczka@dimatics.com">Michal Maczka</a>
- *
+ *
```
michael-o, line 30: Unrelated

wagon-provider-api/src/main/java/org/apache/maven/wagon/authentication/AuthenticationInfo.java
michael-o: Same here

wagon-provider-api/src/main/java/org/apache/maven/wagon/authentication/AuthenticationInfo.java
michael-o: The upstream model does not allow that. Given that other password fields are strings too, this is acceptable, but not ideal.

wagon-provider-api/src/main/java/org/apache/maven/wagon/authentication/AuthenticationInfo.java
michael-o: Same here

wagon-provider-api/src/main/java/org/apache/maven/wagon/authentication/AuthenticationInfo.java
```diff
@@ -51,7 +52,300 @@
...
+    }
+
+    /**
+     *
+     *
+     * The path to the keystore used for authentication purposes, or null
+     *
+     * .
+     *
+     * @return path, named keystore (such as MY) or null
```
michael-o, line 220: Bad example, if at all `Windows-MY`.

wagon-provider-api/src/main/java/org/apache/maven/wagon/repository/Repository.java
```diff
@@ -19,13 +19,12 @@
  * under the License.
  */
 
+import java.io.Serializable;
+import java.util.Properties;
 import org.apache.maven.wagon.PathUtils;
 import org.apache.maven.wagon.WagonConstants;
 import org.codehaus.plexus.util.StringUtils;
 
-import java.io.Serializable;
```
michael-o, line 26: Same here.

wagon-provider-api/src/main/java/org/apache/maven/wagon/repository/Repository.java
```diff
@@ -53,7 +52,7 @@
 
     private String url;
 
-    private RepositoryPermissions permissions;
+    private RepositoryPermissions permissions;
```
michael-o, line 55: traling ws

wagon-provider-api/src/main/java/org/apache/maven/wagon/repository/Repository.java
```diff
@@ -65,7 +64,7 @@
     private String username = null;
 
     private String password = null;
-
+
```
michael-o, line 67: traling ws

wagon-providers/wagon-http-shared/src/main/java/org/apache/maven/wagon/shared/http/AbstractHttpClientWagon.java
```diff
@@ -80,37 +114,11 @@
 import org.apache.maven.wagon.proxy.ProxyInfo;
 import org.apache.maven.wagon.repository.Repository;
 import org.apache.maven.wagon.resource.Resource;
-import org.codehaus.plexus.util.StringUtils;
-
-import javax.net.ssl.HttpsURLConnection;
```
michael-o, line 85: same here

wagon-providers/wagon-http-shared/src/main/java/org/apache/maven/wagon/shared/http/AbstractHttpClientWagon.java
```diff
@@ -382,12 +390,12 @@ private static PoolingHttpClientConnectionManager createConnManager()
         {
             sslConnectionSocketFactory =
                 new SSLConnectionSocketFactory( HttpsURLConnection.getDefaultSSLSocketFactory(), sslProtocols,
-                                                cipherSuites,
+                    cipherSuites,
```
michael-o, line 393: Drop reformat

wagon-providers/wagon-http-shared/src/main/java/org/apache/maven/wagon/shared/http/AbstractHttpClientWagon.java
```diff
@@ -382,12 +390,12 @@ private static PoolingHttpClientConnectionManager createConnManager()
...
                                                 SSLConnectionSocketFactory.BROWSER_COMPATIBLE_HOSTNAME_VERIFIER );
         }
-
+        
         Registry<ConnectionSocketFactory> registry = RegistryBuilder.<ConnectionSocketFactory>create().register( "http",
-                                                                                                                 PlainConnectionSocketFactory.INSTANCE ).register(
+                        PlainConnectionSocketFactory.INSTANCE ).register(
```
michael-o, line 398: Drop reformat

wagon-providers/wagon-http-shared/src/main/java/org/apache/maven/wagon/shared/http/AbstractHttpClientWagon.java
```diff
@@ -593,13 +603,189 @@ private static CloseableHttpClient createClient()
      */
     private BasicAuthScope proxyAuth;
 
+    /**
+     * initializes a custom http client given user specified keystore/truststore
+     * information from settings.xml
+     * see httsp://issues.apache.org/jira/browser/MNG-5583
```
michael-o, line 609: Typo

wagon-providers/wagon-http-shared/src/main/java/org/apache/maven/wagon/shared/http/AbstractHttpClientWagon.java
```diff
@@ -593,13 +603,189 @@ private static CloseableHttpClient createClient()
...
+                    if ( keystore.containsAlias( alias ) ) 
+                    {
+                        if ( !keystore.isKeyEntry( alias ) ) 
+                        {
+                            alias = null;
+                        }
+                    } 
+                    else if ( keystore.containsAlias( alias.toLowerCase() ) ) 
+                    {
+                        alias = alias.toLowerCase();
```
michael-o, line 705: Prone to locale issues.

wagon-providers/wagon-http-shared/src/main/java/org/apache/maven/wagon/shared/http/AbstractHttpClientWagon.java
```diff
@@ -593,13 +603,189 @@ private static CloseableHttpClient createClient()
...
+                        if ( !keystore.isKeyEntry( alias ) ) 
+                        {
+                            alias = null;
+                        }
+                    }
+                    if ( alias == null ) 
+                    {
+                        //TODO this should be I18N
+                        throw new IOException( "key alias not found" );
```
michael-o, line 714: Why is that an IOE?

wagon-providers/wagon-http-shared/src/main/java/org/apache/maven/wagon/shared/http/AbstractHttpClientWagon.java
```diff
@@ -593,13 +603,189 @@ private static CloseableHttpClient createClient()
...
+    
+    
     public void openConnectionInternal()
     {
         repository.setUrl( getURL( repository ) );
 
         credentialsProvider = new BasicCredentialsProvider();
         authCache = new BasicAuthCache();
-
+
```
michael-o, line 788: trailing ws

wagon-providers/wagon-http-shared/src/main/java/org/apache/maven/wagon/shared/http/AbstractHttpClientWagon.java
```diff
@@ -614,11 +800,23 @@ public void openConnectionInternal()
...
+                    httpClientWithCustomSslSocketFactoryCache.put( key, 
+                            initilaizeLocalHttpClientWithCustomSslSocketFactory() );
+                }
+            }
         }
 
         ProxyInfo proxyInfo = getProxyInfo( getRepository().getProtocol(), getRepository().getHost() );
         if ( proxyInfo != null )
         {
+            //TODO add PKI support?
```
michael-o, line 819: This would require two things:

* A proxy which speaks HTTPS as its protocol
* The proxy from above + TLS mutual auth supported

If this cannot be set up with OSS tools no need to add a TODO here. Maybe Apache's mod_proxy might be tried as forwarding proxy.

wagon-provider-api/src/main/java/org/apache/maven/wagon/authentication/AuthenticationInfo.java
spyhunter99: np, someone should tell the netbeans folks then. auto formatter does that. If it bothers you all that much, add it to the check style xml file

spyhunter99: actually i think it was from the maven core mdo > java class generator that did that and i just copied and pasted... no worries

wagon-providers/wagon-http-shared/src/main/java/org/apache/maven/wagon/shared/http/AbstractHttpClientWagon.java
spyhunter99: this was borrowed from tomcat's code base, apparently in certain situational, key aliases are case folded. i've never seen it in the wild either, it's always been case sensitive but i figured there was a reason for this.

wagon-providers/wagon-http-shared/src/main/java/org/apache/maven/wagon/shared/http/AbstractHttpClientWagon.java
spyhunter99: i'm pretty sure it's possible, but i'll remove the todo if you like.

wagon-providers/wagon-http-shared/src/main/java/org/apache/maven/wagon/shared/http/AbstractHttpClientWagon.java
spyhunter99: i can change to SSLInitializationException, would that work?

wagon-providers/wagon-http-shared/src/main/java/org/apache/maven/wagon/shared/http/AbstractHttpClientWagon.java
spyhunter99: what's your preference for addressing this? with everything on one line, it was triggering other checkstyle rules for max length

wagon-provider-api/src/main/java/org/apache/maven/wagon/authentication/AuthenticationInfo.java
spyhunter99: ok i'll remove the reference, but windows-my, while it is obviously a platform specific thing, as is the mac specific version, does work and is (in some cases) the only way to access specific certificates

wagon-provider-api/src/main/java/org/apache/maven/wagon/AbstractWagon.java
spyhunter99: artifact of netbean's fix imports. i'll roll it back

wagon-providers/wagon-http-shared/src/main/java/org/apache/maven/wagon/shared/http/AbstractHttpClientWagon.java
spyhunter99: true, i'll remove it if you want

wagon-providers/wagon-http-shared/src/main/java/org/apache/maven/wagon/shared/http/AbstractHttpClientWagon.java
spyhunter99: ok, i was unsure of the java min version for wagon so i went with the safest route.

spyhunter99:
FYI i'm going to try and add some new unit tests for this to have a more reproducible setup rather than the hacked up sonatype nexus server i'm testing against

wagon-provider-api/src/main/java/org/apache/maven/wagon/repository/Repository.java
```diff
@@ -19,12 +19,12 @@
  * under the License.
  */
 
+import java.util.Properties;
```
mthmulders, line 22: I think you missed one over here.

wagon-providers/wagon-http-shared/src/main/java/org/apache/maven/wagon/shared/http/AbstractHttpClientWagon.java
```diff
@@ -593,6 +610,141 @@ private static CloseableHttpClient createClient()
...
+        String[] cipherSuites = cipherSuitesStr != null ? cipherSuitesStr.split( " *, *" ) : null;
+
+        SSLSocketFactory socketFactory = null;
+        TrustManager[] trustManagers = null;
+        KeyManager[] keyManagers = null;
+        try
+        {
+            if ( authenticationInfo.getTrustStore() != null )
+            {
+                KeyStore keystore = initializeKeyStore( authenticationInfo.getTrustStoreType(),
```
mthmulders, line 634: Conceptually, the variable of type `Keystore` here is a _trust store_, but it's no big deal to me.

spyhunter99:
regarding unit tests, whats the difference between the regular http wagon and the light weight version?

michael-o:
> 
> 
> regarding unit tests, whats the difference between the regular http wagon and the light weight version?

Bad naming. HttpWagon uses Apache HttpClient, LightweightClient uses `HttpURLConnection`.

spyhunter99:
> 
> 
> > regarding unit tests, whats the difference between the regular http wagon and the light weight version?
> 
> Bad naming. HttpWagon uses Apache HttpClient, LightweightClient uses `HttpURLConnection`.

perhaps i asked the wrong question. In what context is the lightweight version used vs the httpwagon version? i'm trying to figure out if i need to (or should) add the pki support mechanisms to the lightweight setup

wagon-provider-api/src/main/java/org/apache/maven/wagon/authentication/AuthenticationInfo.java
spyhunter99: i agree with @michael-o otherwise i would have used char[]

wagon-providers/wagon-http-shared/src/main/java/org/apache/maven/wagon/shared/http/AbstractHttpClientWagon.java
spyhunter99: here's the link
https://github.com/apache/tomcat/blob/a6aad15a470f6f69bcc58c53729dc4e787a536ed/java/org/apache/tomcat/util/net/SSLUtilBase.java#L371

michael-o:
> 
> 
> > > regarding unit tests, whats the difference between the regular http wagon and the light weight version?
> > 
> > 
> > Bad naming. HttpWagon uses Apache HttpClient, LightweightClient uses `HttpURLConnection`.
> 
> perhaps i asked the wrong question. In what context is the lightweight version used vs the httpwagon version? i'm trying to figure out if i need to (or should) add the pki support mechanisms to the lightweight setup

Not in Maven itself. It is a user desicion to swap the HTTP client backend for Wagon. The LightweightWagon has flaws as documented in JIRA.

spyhunter99:
anything else that needs to changed on this PR?

wagon-providers/wagon-http-shared/src/main/java/org/apache/maven/wagon/shared/http/JSSEKeyManager.java
```diff
@@ -0,0 +1,139 @@
...
+import java.security.Principal;
+import java.security.PrivateKey;
+import java.security.cert.X509Certificate;
+
+import javax.net.ssl.SSLEngine;
+import javax.net.ssl.X509ExtendedKeyManager;
+import javax.net.ssl.X509KeyManager;
+
+/**
+ * This was borrowed from Apache Tomcat
```
mthmulders, line 32: Does Tomcat also have a unit test for this class? If so, I think we might want to borrow that as well.

wagon-providers/wagon-http-lightweight/src/test/java/org/apache/maven/wagon/providers/http/TckTest.java
```diff
@@ -20,6 +20,7 @@
  */
 
 import org.apache.maven.wagon.tck.http.GetWagonTests;
+import org.apache.maven.wagon.tck.http.HttpsClientCertGetWagonTests;
```
mthmulders, line 23: I don't see why this is necessary. Could you please explain?

wagon-tcks/wagon-tck-http/src/main/java/org/apache/maven/wagon/tck/http/GetWagonTests.java
```diff
@@ -545,4 +549,9 @@ protected void testErrorHandling( final int code )
             // expected
         }
     }
+
+    protected  AuthenticationInfo getAuthIfNeeded( AuthenticationInfo info ) throws Exception
```
mthmulders, line 553: What is the purpose of this method?

michael-o:
Since this is a massive change, I have to complete the current Wagon release first and then will pick this up in June.

wagon-providers/wagon-http-lightweight/src/test/java/org/apache/maven/wagon/providers/http/TckTest.java
spyhunter99: good catch, i think i had attempted to enable that test profile for the lightweight provider, but considering your explanation of when it's used, it's out of scope anyhow.

wagon-providers/wagon-http-shared/src/main/java/org/apache/maven/wagon/shared/http/JSSEKeyManager.java
spyhunter99: Doesn't look like it. I'll see if I can whip one up. The  purpose of the class is to force a selection for a specific certificate. The default JRE implementation will pick the key pair that matches the server's criteria as part of the hand shake. If you have more than one that matches the criteria but only a specific one will allow access, you're hosed.

spyhunter99: actually i'm glad you brought this up, definitely needs a test case. I shouldn't have copied tomcat's code without reading the details. It's purpose is to select the server certificate to for when a single tomcat server hosts multiple domains all in the same key store. We actually need the opposite, client certificate selection.

PR #68, opened 2020-05-31
WAGON-597 Fix self-assignment and set class field

pzygielo:
@olamy - may I ask for review, please?


michael-o:
Please create a JIRA issue. This is a long standing one I know about and always forget to fix.

PR #70, opened 2020-07-24
Speedup CI builds with hashver-maven-plugin?

avodonosov:
Hi, are you interested to speedup Ci builds?

I'm working on a tool that allows to build and test only submodules affected by the changes (taking the dependency tree into account) - https://github.com/avodonosov/hashver-maven-plugin - and use the maven-wagon project as a testbench for the tool.

This PR demonstrates how to integrate the tool into the project.

CI scripts need to be adjusted too. If you are interested I can help with that.

Best regards,
- Anton


elharo:
probably not is my gut, but if you want to make a case for this I suggest opening a discussion on the dev mailing list.

PR #74, opened 2020-11-24
fix encode URL to preserve trailing slash

volkodava:
In this pull request we have a fix for breaking change introduced by `EncodingUtil.encodeURLToString( String baseUrl, String path )` for URLs with trailing slash in the following commit: https://github.com/apache/maven-wagon/commit/7d1315e47df948d8b17f5c759b20a5b3437cbff7#diff-a0742c473bc4d821caf820463c87956b67da4d07c059d5b6a631e98dd3f7f7b4R93, which has two issues:
- inconsistency between `EncodingUtil.encodeURLToString( String url )` and `EncodingUtil.encodeURLToString( String baseUrl, String path )` in regards to URLs with trailing slash
- affects `org.apache.maven.wagon.shared.http.AbstractHttpClientWagon.resourceExists(...)` API, as URL `https://host:1234/test/demo.zip/` gets normalized to `https://host:1234/test/demo.zip` (with no trailing slash) by `EncodingUtil.encodeURLToString( String baseUrl, String path )` before verify if resource exists

### DIFF
--- a/wagon-providers/wagon-http-shared/src/main/java/org/apache/maven/wagon/shared/http/EncodingUtil.java
+++ b/wagon-providers/wagon-http-shared/src/main/java/org/apache/maven/wagon/shared/http/EncodingUtil.java
@@ -92,7 +92,13 @@ public class EncodingUtil
     {
         String[] pathSegments = path == null ? new String[0] : path.split( "/" );
 
-        return encodeURLToString( baseUrl, pathSegments );
+        String encodedUrl = encodeURLToString( baseUrl, pathSegments );
+        if ( path != null && path.endsWith( "/" ) )
+        {
+            return encodedUrl + "/";
+        }
+
+        return encodedUrl;
     }
 
     /**
--- a/wagon-providers/wagon-http-shared/src/test/java/org/apache/maven/wagon/shared/http/EncodingUtilTest.java
+++ b/wagon-providers/wagon-http-shared/src/test/java/org/apache/maven/wagon/shared/http/EncodingUtilTest.java
@@ -25,8 +25,31 @@ import java.net.MalformedURLException;
 import java.net.URISyntaxException;
 
 public class EncodingUtilTest
-    extends TestCase
-{
+        extends TestCase {
+    public void testEncodeURLWithTrailingSlash()
+    {
+        String baseUrl = "https://host:1234/test";
+        String path = "demo.zip/";
+        String expectedUrl = String.format( "%s/%s", baseUrl, path );
+
+        String encodedURL = EncodingUtil.encodeURLToString( baseUrl, path );
+
+        assertEquals( expectedUrl, encodedURL );
+    }
+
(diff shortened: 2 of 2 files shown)


michael-o:
I need to take a closer look before merging this.

michael-o:
OK, let me get to the breaking change first: The entire old code has been thrown out because it was defective and did not really comply with RFC 3986. HttpClient offers better solutions for this. Now going through your change.

michael-o:
So I have checked RFC 3986, it does not say anything about path ending normalization. Therefore I have checked WHATWG's URL spec: https://url.spec.whatwg.org/#url-serializing. Especially 4.2, append `/` and segment. We need to split the path because it could come from an FS unencoded.  Now what about parsing? If I understand [this](https://url.spec.whatwg.org/#path-state) correctly, the parser remains in path state until a termination or a state change happens. I think 4.2 applies here in the parsing, after the last slash the buffer is empty, but still appended. So `/test/demo.zip/` splits up into:
0: `"test"`
1: `"demo.zip"`
3: `""`

So does the serialization happens. I need to check a few things and ask another expert. @ok2c What is your opinion on?
So you PR is partially correct and the trailing slash should logically be retained, but not unconditionally appended. I will get back to you when my checks are complete.

michael-o:
Please create a JIRA issue for this.

michael-o:
I can confirm that HttpClient's `URLEncodedUtils.parsePathSegments("/wow/"))` does what I have expected: `[wow, ]`.
We need now a split which retains the last empty element.

volkodava:
@michael-o respective JIRA created: https://issues.apache.org/jira/browse/WAGON-602, `encodeURLToString` updated to use `parsePathSegments` from `URLEncodedUtils`. Please find more details at: https://github.com/apache/maven-wagon/pull/74/commits/1b8361402d5613fe9ee8045703e128776761cbee

michael-o:
You have to undo the last commit because `URLEncodedUtils` expects encoded paths.

volkodava:
> You have to undo the last commit because `URLEncodedUtils` expects encoded paths which we don't have.
done

ok2c:
> So does the serialization happens. I need to check a few things and ask another expert. @ok2c What is your opinion on?
> So you PR is partially correct and the trailing slash should logically be retained, but not unconditionally appended. I will get back to you when my checks are complete.

@michael-o Yes, this sounds correct. The parser should be able to differentiate `/blah/blah/` and `/blah/blah` as distinct paths and do not turn one into another. The former should be treated as ['blah', 'blah', ''] sequence and the latter as ['blah', 'blah']

michael-o:
@ok2c Thank you. This is what I have assumed!
@volkodava I have pushed branch WAGON-602 which another test, but removed one test because it was logically wrong. As soon as you add a non-ASCII character your test will fail. I will run the entire test set and them merge into master.

PR #79, opened 2021-12-18
[WAGON-618] Remove HTTP file listing with JSoup

michael-o:
https://issues.apache.org/jira/browse/WAGON-618


hboutemy:
short question: what is the user-visible effect?

longer form: I suppose no user knows that JSoup is used somewhere to list the content of a remote HTTP directory. Until now, it was magically working, sometimes...
Now that the API will always *not* work (consistently), is there a user visible output? something like a warning ("listing files on HTTP server xxx is not supported"), or an exception?

And basic question: does it only affect the getFileList(String) API for HTTP? https://maven.apache.org/wagon/wagon-provider-api/apidocs/org/apache/maven/wagon/Wagon.html#getFileList-java.lang.String-

michael-o:
> short question: what is the user-visible effect?

Exception on a call.

> longer form: I suppose no user knows that JSoup is used somewhere to list the content of a remote HTTP directory. Until now, it was magically working, sometimes... Now that the API will always _not_ work (consistently), is there a user visible output? something like a warning ("listing files on HTTP server xxx is not supported"), or an exception?

Yes, `UnsupportedOperationException` will be thrown with an appropriate message.

> And basic question: does it only affect the getFileList(String) Wagon API for HTTP providers? https://maven.apache.org/wagon/wagon-provider-api/apidocs/org/apache/maven/wagon/Wagon.html#getFileList-java.lang.String-

Yes, only. HTTP. Everything else continues to work.

hboutemy:
nice, I love the clear message, thank you

hboutemy (approved):

michael-o:
Merged into master.

jira-importer:
Resolve #755

PR #83, opened 2022-06-17
[WAGON-627] Maven deploy fails with 401 Unauthorized when using £ in …

michael-o:
…password

When configuring preemptive authentication we need to make sure that it uses
the same character encoding as the custom auth scheme registry. Then characters
are properly encoded into bytes.

This closes #83


olamy (approved):

jira-importer:
Resolve #787

PR #84, opened 2022-06-18
[WAGON-628] Default connect timeout not set when no HttpMethodConfigu…

michael-o:
…ration is available

Set first request config values then have every HTTP method separately apply
its config. They can safely overwrite timeouts even with -1 (indefinite).
There is also no need to explicitly set PUT default config because it is always
set before config is applied.

This closes #84


jira-importer:
Resolve #764

PR #89, opened 2023-04-23
[WAGON-632] Java 8

elharo:
If there's a reason we can't move to Java 8, we can perhaps use Files.copy instead of plexus to copy the directory in FileWagon. Might want to do that anyway. Bugs there seem to be at the root of more than one CI failure.

### DIFF
--- a/.github/workflows/maven-verify.yml
+++ b/.github/workflows/maven-verify.yml
@@ -24,4 +24,4 @@ on:
 jobs:
   build:
     name: Verify
-    uses: apache/maven-gh-actions-shared/.github/workflows/maven-verify.yml@v2
+    uses: apache/maven-gh-actions-shared/.github/workflows/maven-verify.yml@v3
--- a/pom.xml
+++ b/pom.xml
@@ -40,7 +40,7 @@ under the License.
     <maven.test.redirectTestOutputToFile>true</maven.test.redirectTestOutputToFile>
     <slf4jVersion>1.7.36</slf4jVersion>
     <maven.site.path>wagon-archives/wagon-LATEST</maven.site.path>
-    <javaVersion>7</javaVersion>
+    <javaVersion>8</javaVersion>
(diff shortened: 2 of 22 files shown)


michael-o:
For this to happen master needs to be moved to 4.0.0-M1-SNAPSHOT and then this issue assigned to it.

michael-o (changes requested):
For this to happen master needs to be moved to 4.0.0-M1-SNAPSHOT and then this issue assigned to it.

slachiewicz (approved):

HannesWell:
> For this to happen master needs to be moved to 4.0.0-M1-SNAPSHOT and then this issue assigned to it.

Would it be possible to replace the httpcomponent dependencies by the classes from the `java.net.http` packages which was added in Java-11? If yes, would it then be an option to even go to Java-11?

michael-o:
> > For this to happen master needs to be moved to 4.0.0-M1-SNAPSHOT and then this issue assigned to it.
> 
> Would it be possible to replace the httpcomponent dependencies by the classes from the `java.net.http` packages which was added in Java-11? If yes, would it then be an option to even go to Java-11?

This can be a separate module which someone needs to write.

slachiewicz:
LGTM @elharo

elharo:
Tests have not passed.

@michael-o are you OK with this now?

michael-o:
> Tests have not passed.
> 
> @michael-o are you OK with this now?

Yes

elharo:
Let me see if I can get the tests to pass

elharo:
#93 might render this moot

elharo:
@slachiewicz You can merge this if you like once @michael-o dismisses the requested change. However if you prefer to simply merge #93 and close this PR, I think that works too.

slachiewicz:
implicitly fixed by parent pom update https://issues.apache.org/jira/browse/WAGON-634

jira-importer:
Resolve #774

PR #96, opened 2024-02-12
Add Dependabot setup

slachiewicz:


rmannibucau:
Any way to make dependabot respect the commit policy?
Or do we just do it at release time, I'm not convinced it is a good thing from the set up projects which have a broken commit history now.

olamy (approved):

slawekjaranowski (approved):

PR #133, opened 2024-12-26
[WAGON-638] Drop deprecated providers

slachiewicz:


michael-o:
The WebDAV provider can only be dropped when its functionality has been folded into the HTTP one, unless we say we drop WebDAV altogether...

slachiewicz:
good point Michael. This removal will also affect plugin https://github.com/mojohaus/wagon-maven-plugin

michael-o (changes requested):
The WebDAV provider needs to be retained until it is foled.

PR #135, opened 2024-12-27
[WAGON-641] POM cleanup and upgrade misc deps 

slachiewicz:


elharo (approved):

michael-o:
Is it worth to fiddle with 3.x?

slachiewicz:
IMHO why not.  So far it is not clear what should be in 4.x - if it's safe for removing deprecated providers or not.

slachiewicz:
We still have dependency on those artefacts in Maven. https://github.com/apache/maven/blob/master/compat%2Fmaven-compat%2Fsrc%2Fmain%2Fjava%2Forg%2Fapache%2Fmaven%2Frepository%2Flegacy%2FWagonManager.java

wagon-providers/pom.xml
```diff
@@ -52,24 +51,11 @@ under the License.
         <artifactId>httpclient</artifactId>
         <version>4.5.14</version>
       </dependency>
-      <!-- Overwrite transitive dependency version of HttpClient -->
-      <dependency>
-        <groupId>commons-codec</groupId>
-        <artifactId>commons-codec</artifactId>
-        <version>1.16.1</version>
-        <scope>runtime</scope>
-      </dependency>
```
michael-o, line 61: This does not apply anymore?

wagon-providers/pom.xml
```diff
@@ -52,24 +51,11 @@ under the License.
...
         <groupId>org.apache.httpcomponents</groupId>
         <artifactId>httpcore</artifactId>
         <version>4.4.16</version>
       </dependency>
-      <dependency>
-        <groupId>org.apache.sshd</groupId>
-        <artifactId>sshd-core</artifactId>
-        <version>0.8.0</version>
-        <scope>test</scope>
-      </dependency>
```
michael-o, line 72: Not required?

slachiewicz: its required but only in one specific module here https://github.com/apache/maven-wagon/pull/135/files/72c1253cbac6240af6290ef01b40a8027a78f8a3#diff-15f3b17854f455ff3800ee2b45fa961a70bae1876256de66dc9bcd06bc19e1b3R58

wagon-providers/pom.xml
slachiewicz: i see no longer reports of active issues with latest version of httpclient

michael-o (approved):

PR #145, opened 2024-12-27
[WAGON-636] Change location for dependency-reduced-pom.xml

slachiewicz:


elharo (changes requested):
I'm still not convinced that this is the right approach. Why are tests generating output into wagon-providers/wagon-http/dependency-reduced-pom.xml? Why shouldn't this be under target/? If it is in wagon-providers/wagon-http/dependency-reduced-pom.xml will mvn clean fix it?

slachiewicz:
Maybe reporting to maven-shade-plugin / update docs there should be a good way to go?

elharo:
> Maybe reporting to maven-shade-plugin / update docs there should be a good way to go?

Is that what's creating the problem here? I haven't dug into it and am not sure what the root cause is. I'm just reporting the symptoms. 

It's possible this makes sense for some shaded projects, though I tend to doubt that, but definitely not for integration tests.

slachiewicz:
This is not in the integration test but while creating artefacts for wagon-http

elharo:
Whatever's going on the build should not be creating new artifacts in the root directory

slachiewicz:
ok, moved to target subdir

elharo (approved):

PR #166, opened 2025-06-01
Simplify exceptions that fail tests

elharo:


olamy:
use the login button and you will be able to re run a test
I personally find the UI for console output way more convenient for logs etc... and result of unit test are so better to find failing unit test

slachiewicz (approved):

PR #187, opened 2025-06-23
Replace plexus IO methods with JDK methods

elharo:
This reproduces existing dependencies on the default character set which is not always correct. That's going to require another PR.


elharo:
I don't release anything these days, but if we're not going to release HEAD and only release wagon-3.x then we should nuke the main branch and make wagon-3.x the main branch.

slachiewicz (approved):

Issue #680, opened 2021-11-11
[WAGON-613] Switching from Jsch to an updated fork


jira-importer:
**[Matthias Wiedemann](https://issues.apache.org/jira/secure/ViewProfile.jspa?name=mwiede)** opened **[WAGON-613](https://issues.apache.org/jira/browse/WAGON-613?redirect=false)** and commented

Hi, I started a Jsch fork some time ago (https://github.com/mwiede/jsch) and since we got a ticket yesterday, which corresponds to wagon-ssh https://github.com/mwiede/jsch/issues/95 we are wondering, whether you might be interested in switching from original com.jcraft:jsch to our fork. Please have a look.

Thanks for your consideration, kind regards

Matthias

---

**Affects:** 3.4.3

**Issue Links:**
- [WAGON-616](https://issues.apache.org/jira/browse/WAGON-616) Deprecate Wagon SSH Provider

- [WAGON-616](https://issues.apache.org/jira/browse/WAGON-616) Deprecate Wagon SSH Provider

1 votes, 3 watchers

jira-importer:
**[Michael Osipov](https://issues.apache.org/jira/secure/ViewProfile.jspa?name=michael-o)** commented

In general yes, I have discussed the state of JSch mit several other committers several times. We are even considering to throw away JSch and move to another SSH impl, e.g., Apache MINA SSHD. The fact that JSch does not really provide a source repo and their dev process is completely opaque makes this library in most cases usesess, regardless of the current deficiencies.

slachiewicz:
Still current, and the case is stronger now than in 2021.

`wagon-ssh` is unchanged on this front:

```xml
<groupId>com.jcraft</groupId>
<artifactId>jsch</artifactId>
<version>0.1.55</version>          <!-- last released 2018 -->
...

```

The provider page says it is deprecated and will be removed in 4.0.0, and that could be read as a decision against doing anything here. It should not be, because 4.0.0 does not exist:

- the last release is 3.5.3, and this branch is 3.5.4-SNAPSHOT;
- `master` is 4.0.0-M1-SNAPSHOT and has never been tagged;
- both a Maven 3.9.16 distribution and a build of Maven 4.0.0-SNAPSHOT bundle `wagon-*` at **3.5.3**.

So the line that ships is 3.x, it is expected to stay for the time being, and it carries an SSH stack whose upstream stopped releasing eight years ago. That is a maintenance risk independent of whatever 4.0.0 eventually does.

[...]

PR #903, opened 2026-08-08
Move wagon-ssh to the maintained JSch fork and stop a stray key shadowing the agent

### LINKED ISSUE #503: [WAGON-446] SSH agent authentication is no longer working in wagon-ssh 2.10
**[Per Landberg](https://issues.apache.org/jira/secure/ViewProfile.jspa?name=perlan)** opened **[WAGON-446](https://issues.apache.org/jira/browse/WAGON-446?redirect=false)** and commented

After upgrading to wagon-2.10,  SSH Agent authentication doesn't seems to work anymore. I got the following error when trying to publish a site, using maven-site-plugin 3.4:

Using private key: \<PATH TO id_rsa>
: Error uploading site: Cannot connect. Reason: Auth fail -> [Help 1]

I have no server for this site defined in my settings.xml

After some investigation, I suspect that the cause is WAGON-413 since the  logic in AbstractJschWagon was changed to not use the agent if ScpHelper.getPrivateKey() returns a File.

Unfortunate, ScpHelper.getPrivateKey() seems to have a fallback behavior that always returns the id_rsa key (if one exists) even when no matching server element exists in settings.xml and the effect is that the agent is never consulted.

The authentication will then fail since no server element provides a key password.

My current workaround is to set the wagon.privateKeyDirectory System Property to an empty directory. This will prevent the ScpHelper.getPrivateKey() fallback to kick in.

Has anybody else got SSH agent working in combination with existing id_rsa files?

---

**Affects:** 2.10

**Attachments:**
- [wagon-446.patch](https://issues.apache.org/jira/secure/attachment/12766315/wagon-446.patch) (_828 bytes_)

**Issue Links:**
- [WAGON-616](https://issues.apache.org/jira/browse/WAGON-616) Deprecate Wagon SSH Provider

### LINKED ISSUE #901: Upgrade wagon-ssh to the maintained JSch fork (com.github.mwiede:jsch)
### Affected version

3.5.3 and every earlier release, plus `master`.

### Description

`wagon-ssh` depends on `com.jcraft:jsch:0.1.55`, last released on 2018-11-26, and on `com.jcraft:jsch.agentproxy.*:0.0.9` from 2014. Neither has a maintainer. This is not only a hygiene problem — the provider is already broken against current servers and current keys:

- 0.1.55 implements neither `rsa-sha2-256` nor `rsa-sha2-512`, so RSA public key authentication fails against OpenSSH 8.8 and later, which stopped accepting `ssh-rsa` (SHA-1) signatures by default in 2021.
- 0.1.55 cannot read the OpenSSH-v1 private key format, which `ssh-keygen` has emitted by default since OpenSSH 7.8 in 2018.

So a user generating a key today, on a server released any time in the last four years, cannot authenticate with it.

### Proposal

Move to `com.github.mwiede:jsch`, the maintained fork, currently **2.28.6** (2026-07-29). This is what #680 asked for in 2021, from the fork's own author. Recording the evaluation here so the decision has the details in one place.

**Why the fork rather than a different library**

- It keeps the `com.jcraft.jsch` package and the same API. Every type this provider uses is present: `JSch`, `Session`, `ChannelExec`, `ChannelSftp`, `SftpATTRS`, `UserInfo`, `UIKeyboardInteractive`, `HostKey`, `HostKeyRepository`, `IdentityRepository`, `ProxyHTTP`, `ProxySOCKS5`.
- Its base bytecode is Java 8, so it fits the 3.x baseline.
- The JSch coupling here is 7 files and about 1565 lines, all under `wagon-ssh/.../providers/ssh/jsch/`. `wagon-ssh-common` refers to `com.jcraft.jsch.UserInfo` only in a javadoc `@see`, so the shared module has no compile dependency on it.

**Apache MINA SSHD was considered and does not fit today.** It would be the natural choice inside the foundation, and it is already a test dependency of `wagon-ssh-common-test`. But `AbstractJschWagon` supports connecting *through* HTTP and SOCKS5 proxies, and MINA SSHD 2.x has no client-side connect-through-proxy — apache/mina-sshd#481, with a fix deferred to the 3.0 filter chain. Porting onto 2.x would silently drop that for the corporate users most likely to still be deploying over scp. Worth revisiting when 3.0 lands.

`sshj` was not pursued: it costs a full port like MINA and adds a mandatory BouncyCastle dependency.

**Licensing** is not a concern: the fork keeps JSch's BSD/ISC terms, which are ASF Category A. Apache Camel (CAMEL-17835) and Apache Commons VFS (VFS-837) have both already made this switch.

### What the change involves

Not quite a coordinate swap. The `com.jcraft.jsch.agentproxy.*` types the provider imports — `ConnectorFactory`, `Connector`, `RemoteIdentityRepository` — have no counterpart in the fork, which integrates the agent as `SSHAgentConnector`, `AgentConnector` and `AgentIdentityRepository`. That is roughly fifteen lines in `AbstractJschWagon`, and the two `jsch.agentproxy` artifacts go away.

### Behaviour changes to put in the release notes

- The fork disables `ssh-rsa` (SHA-1) by default. Deploying to an old server or appliance that offers nothing newer needs `jsch.server_host_key` and `jsch.client_pubkey` set. Both replace the list wholesale, so name every algorithm to keep, not just the one to add — `-Djsch.server_host_key=ssh-ed25519,rsa-sha2-512,rsa-sha2-256,ssh-rsa`. The OpenSSH `+`/`-` prefix syntax is **not** accepted by these properties; it works only in a config file read through `OpenSSHConfig`.
- Ed25519 keys need Java 15+ at runtime, or BouncyCastle on the classpath.
- Reaching the SSH agent uses a Unix domain socket, which the JDK provides from Java 16. Below that it depends on a helper library being present.
- Host key type negotiation order differs, so a host whose `known_hosts` entry is an old `ssh-rsa` key may need re-accepting.

### Related

- #680 — the original request from the fork's author, which this carries forward.
- #503 — SSH agent authentication not working. Worth doing in the same release, but it is a **separate bug in wagon's own logic** and this change does not fix it.

### Verification before release

The `-Dssh-tests` profile is excluded by default and should be run for this. Those tests use the embedded MINA server and exercise the host key error paths, which match on JSch's exception *message text* — the one place a library swap could break something quietly.

slachiewicz:
Same change as #902, on `master`. The cherry-pick was clean.

Two commits, separable: the first swaps the SSH library, the second fixes an authentication bug that is independent of it.

Fixes #901. Fixes #503. Carries forward #680.

### 1. Move to the maintained JSch fork

`com.jcraft:jsch:0.1.55` was last released in 2018. The provider is not merely unmaintained, it is already broken against current keys — here is 0.1.55 against two keys `ssh-keygen` produced with defaults on this machine:

```
jsch-0.1.55        enc_key   -> JSchException: invalid privatekey
                   plain_key -> JSchException: invalid privatekey
com.github.mwiede  enc_key   -> OK
                   plain_key -> OK
```

It cannot read the OpenSSH-v1 format `ssh-keygen` has emitted by default since OpenSSH 7.8, encrypted or not, and it has no `rsa-sha2-*`, so RSA authentication fails against OpenSSH 8.8 and later anyway.

`com.github.mwiede:jsch:2.28.6` keeps the `com.jcraft.jsch` package, so the API is unchanged. Its base bytecode is Java 8.

Two things needed care rather than a straight coordinate swap:

**The agent.** The `jsch.agentproxy` artifacts have no counterpart, and their `ConnectorFactory` used to pick between agent kinds. The choice is now explicit and still covers all three the old stack could reach — `SSH_AUTH_SOCK`, Win32 OpenSSH, Pageant. One capability does narrow: reaching the `SSH_AUTH_SOCK` agent needs a Unix domain socket, which the JDK provides only from Java 16, and the fork declares no dependencies, so on Java 8–15 that particular agent is out of reach without a helper library on the class path. The other two need nothing. An agent holding no identities is now skipped rather than installed as an empty repository.

**Passphrases.** `getPrivateKey()` substitutes an empty passphrase when the settings name none. The old JSch discarded the result of `setPassphrase`, so an encrypted key stayed encrypted and was unlocked later; the fork rejects it instead. Left alone, that would have failed *every* encrypted key. Empty is now translated back to none before the key is added:

```
enc_key,   settings gave no passphrase -> accepted
plain_key, settings gave no passphrase -> accepted
enc_key,   correct passphrase          -> accepted
enc_key,   wrong passphrase            -> rejected
```

### 2. Stop a stray key file shadowing the agent (#503)

The agent was consulted only when no key file was found, and `getPrivateKey()` finds one whenever no password is configured. On any machine with a key in `~/.ssh` the agent was therefore unreachable — which is why the reported workaround was to point `wagon.privateKeyDirectory` at an empty directory.

The order is now: a key named in `settings.xml`, then the agent, then a key file that merely happens to be in `~/.ssh`. Naming a key keeps today's behaviour.

**This is a behaviour change to note in the release notes.** A reachable agent holding identities now outranks a discovered key file, so someone whose agent holds unrelated keys while the needed key sits unloaded in `~/.ssh` must load it or name it. Combining both is not available to us: adding a file identity on top of an agent repository makes JSch push that key *into* the user's agent.

Also in this commit, key discovery looks for `id_ed25519`, `id_ecdsa`, `id_rsa` instead of `id_dsa` then `id_rsa`. That code is in `ScpHelper`, so it also changes which key `wagon-ssh-external` passes to `ssh -i`.

### Verification, and its limits

Modules build, `spotless:check` clean, default test run green.

**The change is not verified by the project's own SSH tests, because those do not run.** The `-Dssh-tests` profile is excluded by default, and on the unmodified branch point it gives 112 tests / 94 errors — `ComponentLookupException: Unable to lookup component 'org.apache.maven.wagon.Wagon', roleHint: scp`, i.e. plexus test wiring rather than anything SSH. My branch gives exactly the same 112 / 94, so this change neither helps nor hurts it, but it means the embedded-MINA-server suite that would genuinely exercise a library swap is unavailable. That harness deserves fixing before this is relied on, and it is worth doing regardless of this PR.

What was verified directly is the key-handling matrix above, run against both libraries.

Not yet done, and worth someone's hands before release: a real `site-deploy` over scp and sftp against a current OpenSSH, on both Java 11 and Java 17, and one Windows check of the agent path.

### DIFF
--- a/wagon-providers/wagon-ssh-common/src/main/java/org/apache/maven/wagon/providers/ssh/ScpHelper.java
+++ b/wagon-providers/wagon-ssh-common/src/main/java/org/apache/maven/wagon/providers/ssh/ScpHelper.java
@@ -102,6 +102,13 @@ public class ScpHelper {
         return privateKey;
     }
 
+    /**
+     * The key types <code>ssh-keygen</code> produces, most recent first. <code>id_dsa</code> is not among
+     * them: ssh-dss has been disabled by default in OpenSSH for years, so a DSA key is the one least likely
+     * to be accepted by the server we are about to reach.
+     */
+    private static final String[] PRIVATE_KEY_NAMES = {"id_ed25519", "id_ecdsa", "id_rsa"};
+
     private static File findPrivateKey() {
         String privateKeyDirectory = System.getProperty("wagon.privateKeyDirectory");
 
@@ -109,16 +116,14 @@ public class ScpHelper {
             privateKeyDirectory = System.getProperty("user.home");
         }
 
-        File privateKey = new File(privateKeyDirectory, ".ssh/id_dsa");
-
-        if (!privateKey.exists()) {
-            privateKey = new File(privateKeyDirectory, ".ssh/id_rsa");
-            if (!privateKey.exists()) {
-                privateKey = null;
+        for (String name : PRIVATE_KEY_NAMES) {
+            File privateKey = new File(privateKeyDirectory, ".ssh/" + name);
+            if (privateKey.exists()) {
+                return privateKey;
             }
         }
 
-        return privateKey;
+        return null;
     }
 
     public static void createZip(List<String> files, File zipName, File basedir) throws IOException {
--- a/wagon-providers/wagon-ssh-common/src/main/java/org/apache/maven/wagon/providers/ssh/ScpHelper.java
+++ b/wagon-providers/wagon-ssh-common/src/main/java/org/apache/maven/wagon/providers/ssh/ScpHelper.java
@@ -23,6 +23,8 @@ import java.io.FileInputStream;
 import java.io.FileNotFoundException;
 import java.io.FileOutputStream;
 import java.io.IOException;
+import java.security.KeyFactory;
+import java.security.NoSuchAlgorithmException;
 import java.util.List;
 import java.util.zip.ZipEntry;
 import java.util.zip.ZipOutputStream;
@@ -109,6 +111,41 @@ public class ScpHelper {
      */
     private static final String[] PRIVATE_KEY_NAMES = {"id_ed25519", "id_ecdsa", "id_rsa"};
 
+    /**
+     * The same names, with Ed25519 demoted to last. Used when this runtime cannot do Ed25519, so that a key
+     * it can actually use wins -- while still falling back to <code>id_ed25519</code> when that is the only
+     * key present, which is the right answer for wagon-ssh-external, where the host's own <code>scp</code>
+     * does the cryptography and the JVM's capabilities do not apply.
+     */
+    private static final String[] PRIVATE_KEY_NAMES_WITHOUT_ED25519 = {"id_ecdsa", "id_rsa", "id_ed25519"};
+
+    /**
+     * The names to look for, in order. Package-private so both orderings can be tested on any runtime.
+     */
+    static String[] preferredPrivateKeyNames(boolean ed25519Available) {
+        return ed25519Available ? PRIVATE_KEY_NAMES : PRIVATE_KEY_NAMES_WITHOUT_ED25519;
+    }
+
+    /**
+     * Whether this runtime can use an Ed25519 key. The JDK grew EdDSA in Java 15; before that a provider
+     * such as Bouncy Castle has to supply it, which JSch will use when it is on the class path.
+     */
+    private static boolean isEd25519Available() {
+        try {
+            KeyFactory.getInstance("Ed25519");
+            return true;
+        } catch (NoSuchAlgorithmException e) {
+            // not in this JDK; a provider may still supply it
+        }
+
+        try {
+            Class.forName("org.bouncycastle.jce.provider.BouncyCastleProvider");
+            return true;
+        } catch (ClassNotFoundException e) {
+            return false;
+        }
+    }
+
     private static File findPrivateKey() {
         String privateKeyDirectory = System.getProperty("wagon.privateKeyDirectory");
 
@@ -116,7 +153,7 @@ public class ScpHelper {
             privateKeyDirectory = System.getProperty("user.home");
         }
 
-        for (String name : PRIVATE_KEY_NAMES) {
+        for (String name : preferredPrivateKeyNames(isEd25519Available())) {
             File privateKey = new File(privateKeyDirectory, ".ssh/" + name);
             if (privateKey.exists()) {
                 return privateKey;
--- a/wagon-providers/wagon-ssh-common/src/test/java/org/apache/maven/wagon/providers/ssh/ScpHelperTest.java
+++ b/wagon-providers/wagon-ssh-common/src/test/java/org/apache/maven/wagon/providers/ssh/ScpHelperTest.java
@@ -1,0 +1,57 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.maven.wagon.providers.ssh;
+
+import java.util.Arrays;
+
+import org.junit.Test;
+
+import static org.junit.Assert.assertArrayEquals;
+import static org.junit.Assert.assertTrue;
+
+/**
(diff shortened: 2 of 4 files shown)


slachiewicz:
Updated, and the headline is that this is now verified rather than asserted.

**The harness came back.** #904 revives the embedded SSH tests, which had not run in over a decade. Running this branch's two commits on top of it: **42 tests, 0 failures** against a real embedded MINA sshd server. That replaces the "unverified by the project's own tests" caveat this PR opened with. #904 should go in first. Same change as #902.

**It immediately caught a fatal bug in this branch, now fixed.** My agent lookup listed the three connector kinds in a `static final` array:

```java
private static final AgentConnectorFactory[] AGENT_CONNECTORS = {
    SSHAgentConnector::new, WindowsSSHAgentConnector::new, PageantConnector::new
};
```

`PageantConnector` needs JNA, and the fork declares no dependencies, so class initialisation of `ScpWagon` died with `NoClassDefFoundError: com/sun/jna/Pointer` wrapped in `ExceptionInInitializerError`. **`wagon-ssh` would not have loaded at all.** The default test run never instantiates a provider, so it went unnoticed; the embedded harness instantiates one on the first test and failed 24 of 42.

Each connector is now both constructed and used inside its own guard that catches `LinkageError`, so a missing class is one skipped agent rather than a failure to load. Pageant therefore still works if JNA is on the class path and is skipped quietly otherwise. Re-run: 42 of 42.

That is a good argument for #904 on its own merits — this defect was invisible to everything else in the build.

slachiewicz:
Ordering note, now that #904 has merged to `wagon-3.x`: the master equivalent of that
harness did not exist, so this PR had nothing to be verified against. #915 forward-ports
it. **#915 should go in before this one**, mirroring #904 -> #902 on the 3.x side.

Once #915 is in I will rebase this and re-run `-Dssh-tests -Dssh-embedded=true` so the
master half gets the same 42-test verification the 3.x half got.

slachiewicz:
Rebased onto master and brought over the two fixes from #902, now that #915 has landed and
there is a harness to check them against.

**The agent guard.** The constructor references defeated it. A constructor reference resolves
its class when the reference is *evaluated*, which happens at the call site in
`agentIdentityRepository()` — outside the `try` in `identitiesFrom`. So a missing connector
class threw `NoClassDefFoundError` straight past the catch, which is what the javadoc claimed
could not happen:

```
methodref: ESCAPED the guard: NoClassDefFoundError
lambda   : caught: NoClassDefFoundError
```

**The key preference.** master targets Java 8 as well, and the JDK only grew EdDSA in Java 15;
before that JSch needs a provider such as Bouncy Castle, which is not a dependency. The base
`SignatureEdDSA` in the JSch jar says so outright — `"SignatureEdDSA requires Java15+."` — so a
stray `id_ed25519` would be selected and fail even with a usable `id_rsa` beside it. Demoted
rather than skipped, so a lone `id_ed25519` is still found; three unit tests cover both
orderings on any JDK.

**Verified against the suite this time**, which is the whole point of taking #915 first:
`-Dssh-tests -Dssh-embedded=true` gives **42 tests, 0 failures** against a real embedded MINA
sshd, and the full reactor is green.

One note for review: I applied the `ScpHelper` change by hand rather than copying the file from
the 3.x branch. Copying it dragged in three things master had already cleaned up — `IOUtil.close`
in place of try-with-resources, some resurrected commented-out `executeCommand` lines, and a
reverted comment typo fix. The diff here is the Ed25519 change and nothing else.

slachiewicz:
not recomended way for 4.x

PR #905, opened 2026-08-08
[wagon-3.x] Add user, developer and HTTP configuration guides

### LINKED ISSUE #190: [WAGON-3] Development/Testing guide
**[Jason van Zyl](https://issues.apache.org/jira/secure/ViewProfile.jspa?name=jvanzyl)** opened **[WAGON-3](https://issues.apache.org/jira/browse/WAGON-3?redirect=false)** and commented

Describe how to develop wagons, and describe how to use the WagonTestCase to simplify the testing of individual Wagons.

---

**Affects:** 1.0-alpha-2

### LINKED ISSUE #191: [WAGON-1] Create user guide/Guide for the impatient programmer (gipper)
**[Jason van Zyl](https://issues.apache.org/jira/secure/ViewProfile.jspa?name=jvanzyl)** opened **[WAGON-1](https://issues.apache.org/jira/browse/WAGON-1?redirect=false)** and commented

Create a simple user guide using the wagon POJO facade.

---

**Affects:** 1.0-alpha-2

### LINKED ISSUE #490: [WAGON-425] write documentation for HTTP auth configuration
**[Hervé Boutemy](https://issues.apache.org/jira/secure/ViewProfile.jspa?name=hboutemy)** opened **[WAGON-425](https://issues.apache.org/jira/browse/WAGON-425?redirect=false)** and commented

seems like the example configuration written in WAGON-422 deserves at least some publication to show how such advanced configuration is possible

ideally, it should be accompanied with some explanations and links (like https://hc.apache.org/httpcomponents-client-4.5.x/index.html IIUC)

---

**Affects:** 2.8

**Issue Links:**
- [WAGON-422](https://issues.apache.org/jira/browse/WAGON-422) HTTP wagon AuthScope is not definable from settings

slachiewicz:
Three guides that have been requested for a long time.

Closes #191 (WAGON-1, open since 2004). Closes #190 (WAGON-3, open since 2004). Closes #490 (WAGON-425, open since 2014).

| Page | Covers |
|---|---|
| `user-guide.md` | What a Wagon is; provider → role hint → URL scheme for all eight providers; obtaining one as a build extension or by lookup; the connect/get/put/disconnect lifecycle with working code; a capability matrix; `Repository` URL parsing including credentials being stripped from `getUrl()`; `AuthenticationInfo`; proxies and `nonProxyHosts` globbing; the two timeouts; streaming versus file transfer; listeners, including that `transferProgress` runs synchronously on the transfer thread; the exception table with the actual message wording |
| `developer-guide.md` | The SPI; what `AbstractWagon` provides and the two methods a provider must supply; `StreamWagon`'s contract and when it does not fit; both registration mechanisms — the `@plexus.component` tag and a hand-written `components.xml`; `wagon-provider-test` and what you inherit from `WagonTestCase`; the HTTP TCK, its use-case ids, and how to wire it in and mark a case unsupported; running the suites |
| `http-configuration.md` | The `<server><configuration>` block for wagon-http; `httpConfiguration` and the six `HttpMethodConfiguration` properties; the merge semantics and their consequences; both XML spellings Plexus accepts; all 17 recognised `params` keys mapped to their `RequestConfig` setters; auth schemes, preemptive auth, scope overrides, proxy and NTLM; the deprecated `httpHeaders`; 429 backoff; the system-property reference |

Everything is grounded in the source. Where something could not be verified from this repository it was left out rather than guessed — noted below.

`mvn site` builds. The rendered HTML was checked rather than just the exit code, which caught one real defect: Doxia generates heading ids as `What_a_Wagon_is`, not the kebab-case GitHub form, so the cross-page anchors were dead until corrected.

### Deliberately not covered

How `settings.xml` populates `ProxyInfo.ntlmHost`/`ntlmDomain` (that mapping lives in Maven core), how Resolver's wagon transport picks a provider (out of this repository), and which of wagon-http or wagon-http-lightweight wins when both are on the class path (the container decides, not Wagon).

### Things found in the code while writing this

Documented where they affect the reader, not fixed here, and worth separate issues:

- **`usePreemptive` never survives a merge.** `HttpMethodConfiguration.copy()` does not copy it and `ConfigurationUtils.merge()` never applies it, so whenever both `<all>` and the block for a method are present the effective value is always `false`. `HttpWagonPreemptiveTest` passes only because it sets `<all>` alone.
- **NTLM against the target server cannot work.** `NTCredentials` is only ever constructed for a proxy, so the registered NTLM scheme has nothing usable to offer an NTLM-protected repository. This is the most likely way a reader of WAGON-425 would be misled.
- **PUT is always preemptively authenticated**, regardless of configuration; the source carries `// FIXME Perform only when preemptive has been configured`.
- **Per-method timeouts cannot be set back to the default value**, because the merge only takes them when they differ from `Wagon.DEFAULT_*`.
- `BasicAuthScope`'s javadoc documents `/server/proxyBasicAuth`, which matches no field; the field is `proxyAuth`.
- `README.md` tells contributors to run `mvn -Prun-its verify`. There is no `run-its` profile anywhere in this repository.
- `wagon-tcks/wagon-tck-http/sample-tck-consumer` is in no `<modules>`, still 1.0-SNAPSHOT against decade-old dependencies, and is the only thing resembling "how to consume the TCK". The guide documents the real wiring from `wagon-http`'s `TckTest` instead.

The four provider pages carrying the "removed in version 4.0.0" notice were left alone. The new pages describe what exists and do not repeat that claim.

### DIFF
--- a/src/site/markdown/developer-guide.md
+++ b/src/site/markdown/developer-guide.md
@@ -393,6 +393,24 @@ A provider with no `WagonTestCaseConfigurator` configuration at all will fail to
 one up; the configurator component itself must be declared even if
 `<useCaseConfigs>` is empty.
 
+### Run the TCK classes, do not subclass them from your own package
+
+`HttpWagonTests.isSupported()` and `HttpWagonTests.initTest(...)` recover the use-case
+id by walking the current stack and taking the last method name seen before the first
+frame whose class name does not start with `getClass().getPackage().getName()` — the
+package of the *runtime* class. That works for `GetWagonTests` and
+`HttpsGetWagonTests`, which live in `org.apache.maven.wagon.tck.http` alongside the
+frame they are looking for.
+
+It does not work if you subclass a TCK test class from your own package. The very
+first frame examined is then already outside your package, the walk stops with no
+method name, and every test logs `Cannot run test: null` and is treated as
+unsupported — so the suite passes without having tested anything.
+
+Wire the TCK up the way `wagon-http` does: run the TCK classes as they are, through a
+`@Suite.SuiteClasses` aggregator, and express everything provider-specific through
+`<useCaseConfigs>` rather than through a subclass.
+
 ## Running the tests
 
 The full build:
--- a/src/site/markdown/developer-guide.md
+++ b/src/site/markdown/developer-guide.md
@@ -1,0 +1,450 @@
+---
+title: Development and Testing Guide
+date: 2026-08-08
+---
+
+<!--
+Licensed to the Apache Software Foundation (ASF) under one
+or more contributor license agreements.  See the NOTICE file
+distributed with this work for additional information
+regarding copyright ownership.  The ASF licenses this file
+to you under the Apache License, Version 2.0 (the
+"License"); you may not use this file except in compliance
+with the License.  You may obtain a copy of the License at
+
+http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing,
+software distributed under the License is distributed on an
+"AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+KIND, either express or implied.  See the License for the
+specific language governing permissions and limitations
+under the License.
+-->
+
+# Development and Testing Guide
+
+This guide is for someone writing or testing a Wagon provider — either a new one, or
+one of those in this repository. If you only want to *use* a Wagon, read the
+[User Guide](./user-guide.html).
+
+## The modules
+
+| Module | Contents |
+| --- | --- |
+| `wagon-provider-api` | the `Wagon` interface, the base classes, events, exceptions |
+| `wagon-provider-test` | reusable abstract JUnit test cases every provider is expected to pass |
+| `wagon-providers/*` | the providers shipped with Wagon |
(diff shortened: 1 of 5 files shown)


slachiewicz:
Second commit added, and it turned up a code defect worth knowing about before review.

Writing the corrected `BasicAuthScope` javadoc meant actually reading the realm branch, and both arms of its `if`/`else` assign the same thing — so `<realm>ANY</realm>` sets the literal string `ANY` rather than `AuthScope.ANY_REALM`, while host and port next to it map `ANY` correctly. Filed as #910. The code is untouched here; fixing it is a behaviour change and belongs in its own PR.

That also means this PR's first commit had it wrong: `http-configuration.md` said each of host, port and realm accepts `ANY` to mean "match anything". The second commit corrects that to describe the asymmetry, keeps the all-`ANY` example with a note that it only works through the early short circuit, and adds an example showing that widening host and port means omitting realm.

Also in this commit: a developer-guide subsection on running the TCK classes rather than subclassing them from your own package. The use-case id is recovered by walking the stack for the last method before the first frame outside `getClass().getPackage().getName()`, so a consumer subclassing from elsewhere gets `Cannot run test: null` on every case, each treated as unsupported — a suite that goes green without testing anything.

Site still builds; the new heading and the corrected text were checked in the rendered HTML.

PR #928, opened 2026-08-17
Migrate the standalone wagon-http tests to JUnit 5

slachiewicz:
Continues `a8cafa27` with the three tests in `wagon-http` that do not extend the bases published from
`wagon-provider-test`. Those bases stay on `PlexusTestCase`: external providers inherit their test
methods by the `testXxx()` convention, so moving them to `@Test` would leave any consumer on JUnit 4
running none of the inherited tests, with a green build.

`wagon-http` keeps `junit` and gains `junit-vintage-engine` alongside `junit-jupiter`. `HttpWagonTest`,
`HttpsWagonTest` and both Preemptive variants inherit from the published `HttpWagonTestCase` — 58 tests
each — and surefire provisions the jupiter engine but not the vintage one. No versions are needed;
`maven-parent` 49 imports `junit-bom`.

Nothing here needed `assertThrows` or Hamcrest work: no `@Test(expected=…)`, no `@Test(timeout=…)`, no
`@Ignore`, no `assertThat`, and no message-carrying assertions whose argument order would flip.

Verified: `mvn test` on `wagon-http` before and after, comparing per class. 295 test cases before, 295
after; the same two `proxied` tests skipped in both.

| | before | after |
|---|---|---|
| AbstractHttpClientWagonTest | 4 | 4 |
| BasicAuthScopeTest | 5 | 5 |
| HttpClientWagonTest | 3 | 3 |
| HttpWagonTest / HttpsWagonTest | 58 / 58 | 58 / 58 |
| HttpWagonPreemptiveTest / HttpsWagonPreemptiveTest | 58 / 58 | 58 / 58 |
| HttpWagonErrorTest / HttpWagonTimeoutTest | 5 / 4 | 5 / 4 |
| HugeFileDownloadTest / TckTest | 2 / 40 | 2 / 40 |

Worth knowing if you check this yourself: once the platform provider is in play, the `tests=`
attribute surefire writes for `TckTest` reads 20 rather than 40, because that class is a JUnit 4
`@RunWith(Suite.class)` aggregating two suites. The report still carries all 40 `<testcase>` elements
and both suites run — counting the attribute rather than the elements makes it look like twenty tests
were lost.

### DIFF
--- a/wagon-providers/wagon-http/pom.xml
+++ b/wagon-providers/wagon-http/pom.xml
@@ -70,6 +70,17 @@ under the License.
       <scope>test</scope>
     </dependency>
     <dependency>
+      <groupId>org.junit.jupiter</groupId>
+      <artifactId>junit-jupiter</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <!-- HttpWagonTest and its subclasses inherit JUnit 4 tests from the published HttpWagonTestCase -->
+    <dependency>
+      <groupId>org.junit.vintage</groupId>
+      <artifactId>junit-vintage-engine</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
       <groupId>junit</groupId>
       <artifactId>junit</artifactId>
       <scope>test</scope>
--- a/wagon-providers/wagon-http/src/test/java/org/apache/maven/wagon/providers/http/AbstractHttpClientWagonTest.java
+++ b/wagon-providers/wagon-http/src/test/java/org/apache/maven/wagon/providers/http/AbstractHttpClientWagonTest.java
@@ -37,13 +37,13 @@ import org.apache.http.impl.client.CloseableHttpClient;
 import org.apache.http.impl.client.DefaultHttpRequestRetryHandler;
 import org.apache.http.impl.execchain.RedirectExec;
 import org.apache.http.impl.execchain.RetryExec;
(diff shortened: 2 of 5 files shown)


slachiewicz:
Added a second commit: `HugeFileDownloadTest` also comes off `PlexusTestCase`. It belongs here rather
than in its own PR because it needs the `junit-jupiter` declaration this PR adds — on its own it fails
to compile with `package org.junit.jupiter.api does not exist`.

The container was only supplying `Wagon.ROLE` `"http"`, which is `HttpWagon`, and
`AbstractHttpClientWagon` declares no injected collaborators, so `new HttpWagon()` is equivalent to the
lookup. `getBasedir()` becomes a local method with the semantics `PlexusTestCase` gave it, so the four
call sites are unchanged. Dropping the base class also removed the inherited `assertTrue`, hence the
static import.

`ScpWagonWithProxyTest` is deliberately not included. `AbstractJschWagon` has three `@Inject` fields —
`knownHostsProvider`, `interactiveUserInfo` and one more — so instantiating it directly would leave
them null and the test would keep passing while exercising something different. Migrating it needs a
real container, which means introducing `plexus-testing`; that is a dependency decision and should be
its own change.

Verification is unchanged: 295 `<testcase>` elements before and after, 2 skipped in both, per class
identical, with `HugeFileDownloadTest` still at 2.

PR #929, opened 2026-08-17
Migrate the standalone wagon-http tests to JUnit 5

slachiewicz:
Backport of #928. Only tests that do not extend the bases published from `wagon-provider-test` move;
those bases stay on `PlexusTestCase`, since external providers inherit their test methods by the
`testXxx()` convention and would silently run none of them.

`wagon-http` keeps `junit` and gains `junit-vintage-engine` alongside `junit-jupiter`. `HttpWagonTest`,
`HttpsWagonTest` and both Preemptive variants inherit from the published `HttpWagonTestCase` — 58 tests
each — and surefire provisions the jupiter engine but not the vintage one.

`BasicAuthScopeTest` and `HttpClientWagonTest` are byte-identical to master's pre-migration state, so
master's reviewed result is taken verbatim. `AbstractHttpClientWagonTest` has drifted here and needed
two things master never faced: JUnit 5 has no `Assertions.assertThat`, so its four uses move to
`org.hamcrest.MatcherAssert` (Hamcrest stays available through `junit` 4.13.2, which this module keeps);
and `@Ignore` becomes `@Disabled` on a method that carries no `@Test` and so never ran under either
framework.

### Verification

`mvn test` on `wagon-http`, comparing `<testcase>` elements per class rather than the `tests=`
attribute: **295 before, 295 after**, same per class, 2 skipped in both.

Both runs end with one error, and it is not the same test:

| | error |
|---|---|
| before | `HttpsWagonTest.testSecuredPutUnauthorized` |
| after | `HttpWagonTest.testProxiedRequest` |

That is order-dependent interference between these classes, which share a static `HttpClient` and proxy
system properties, not something this change introduces. Run on its own, `HttpWagonTest` is 58 tests and
0 errors on **both** trees — unmigrated and migrated. The engine switch changes execution order, so a
different order-dependent test is the one that loses. Worth knowing when reading CI here; it is a
pre-existing fragility in these tests and is not addressed by this PR.

One reporting quirk if you re-run this yourself: once the platform provider is active, the `tests=`
attribute surefire writes for `TckTest` reads 20 rather than 40, because that class is a JUnit 4
`@RunWith(Suite.class)` aggregating two suites. All 40 `<testcase>` elements are present and both suites
run — counting the attribute makes it look like twenty tests vanished.

### DIFF
--- a/wagon-providers/wagon-http/pom.xml
+++ b/wagon-providers/wagon-http/pom.xml
@@ -70,6 +70,17 @@ under the License.
       <scope>test</scope>
     </dependency>
     <dependency>
+      <groupId>org.junit.jupiter</groupId>
+      <artifactId>junit-jupiter</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <!-- HttpWagonTest and its subclasses inherit JUnit 4 tests from the published HttpWagonTestCase -->
+    <dependency>
+      <groupId>org.junit.vintage</groupId>
+      <artifactId>junit-vintage-engine</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
       <groupId>junit</groupId>
       <artifactId>junit</artifactId>
       <scope>test</scope>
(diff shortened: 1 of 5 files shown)


slachiewicz:
Added the second commit from #928: `HugeFileDownloadTest` comes off `PlexusTestCase`. This branch's
copy was byte-identical to master's before its migration, so master's result is taken verbatim.

The container there was only supplying `Wagon.ROLE` `"http"`, which is `HttpWagon`, and
`AbstractHttpClientWagon` declares no injected collaborators, so a plain constructor is equivalent.

`ScpWagonWithProxyTest` is deliberately not included: `AbstractJschWagon` has three `@Inject` fields,
so instantiating it directly would leave them null and the test would keep passing while exercising
something different. That one needs a real container, i.e. `plexus-testing`, which should be its own
change.

Verification after this commit: 295 `<testcase>` elements, 2 skipped, per class identical, and this
run finished with **no** errors at all — the `testProxiedRequest` error noted in the description did
not reappear, which is consistent with it being the order-dependent interference described there
rather than anything introduced here.

PR #936, opened 2026-08-18
Migrate the published test base classes off PlexusTestCase

slachiewicz:
Completes the JUnit 5 move for the test base classes that wagon publishes: `WagonTestCase`,
`CommandExecutorTestCase`, `HttpWagonTestCase` and `KnownHostsProviderTestCase`. These are
`src/main/java` in released artifacts, so the change reaches consumers who extend them —
which drove three decisions worth a reviewer's attention.

**`plexus-testing` is a compile-scope dependency of `wagon-provider-test` and
`wagon-ssh-common-test`.** `@PlexusTest` sits on published classes, so a consumer extending
`WagonTestCase` needs it on their own compile path; test scope would not propagate. Its version
moves to the root `dependencyManagement` alongside the other third-party versions, since
`wagon-ssh` already declared it and this change would otherwise make three copies.

**Six members disappeared with `PlexusTestCase` and are replaced rather than dropped**, so
subclasses outside this repo keep compiling: `getTestFile`, `getTestPath` and `getBasedir`
delegate to `PlexusExtension`; `getName()` is served by a `TestInfo` callback kept separate from
`setUp()` — eight subclasses override `setUp()` and call `super.setUp()`, and widening that
signature would break every one of them.

**`runTest()` and the `testSkipped` flag are gone, replaced by `assumeTrue`.** This changes a CI
signal: on a machine without `cvs`, the old mechanism reported those wagon-scm tests as
*passing*, and they are now reported as *skipped* (0 → 14 locally). Same tests, honest status.

**Checkstyle `FileLength` is suppressed for `HttpWagonTestCase`** via a new
`src/config/checkstyle-suppressions.xml`. One `@Test` per method took it one line past the
inherited 2000-line limit. Neither `@SuppressWarnings` nor a `CHECKSTYLE_OFF` comment can
suppress it — the reasons are in a comment in that file. Splitting the class is a separate change.

Verified: per-module `<testcase>` element counts in the surefire XML are unchanged against
`f289bdd5` across all eight provider modules (22/23/295/214/42/0/0/295), with no errors or
failures.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -196,6 +196,11 @@ under the License.
       </dependency>
       <dependency>
         <groupId>org.codehaus.plexus</groupId>
+        <artifactId>plexus-testing</artifactId>
+        <version>2.1.0</version>
+      </dependency>
+      <dependency>
+        <groupId>org.codehaus.plexus</groupId>
         <artifactId>plexus-interactivity-api</artifactId>
         <version>1.5.1</version>
       </dependency>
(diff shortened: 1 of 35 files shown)


slachiewicz:
The `wagon-3.x` counterpart is #937. It is not a clean lift — 18 of the 35 files had drifted between the lines, so it carries five deliberate deviations, listed in its description.

